PR #2, opened 2017-02-14
Add monitoring.proto

ericgribkoff:


grpc/instrumentation/v1alpha/monitoring.proto
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+}
+
+// Canonical RPC stats exported by gRPC.
+message CanonicalRpcStats {
+  // Wrapper combining View and ViewDescriptor.
+  message View {
+    google.instrumentation.View view = 1;
+    google.instrumentation.ViewDescriptor view_descriptor = 2;
```
a-veitch: You might want to also have the MeasurementDescriptor here as well.

Nit: I would also put the descriptors before the view.

ericgribkoff: Done. Inclusion of MeasurementDescriptor makes sense. This also led me to remove the `_in_usec` suffix from the time-based metrics. Some of the time measurement descriptors from `RpcConstants` are in milliseconds, others are in usec, and it makes sense to just return the existing measurement descriptors rather than tie each name to a specific unit.

@zhangkun83 - Previously you had suggested dropping units like `usec` from the names in `CanonicalRpcStats`. Can you verify this is the correct approach?

zhangkun83: Yes, this looks good.

grpc/instrumentation/v1alpha/monitoring.proto
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+}
+
+// Canonical RPC stats exported by gRPC.
+message CanonicalRpcStats {
+  // Wrapper combining View and ViewDescriptor.
+  message View {
+    google.instrumentation.ViewDescriptor view_descriptor = 1;
+    google.instrumentation.MeasurementDescriptor measurement_descriptor = 2;
```
a-veitch: I would move measurement_descriptor before view_descriptor - that way everything logically follows the dependency order.

PR #3, opened 2017-03-07
Add StatsRequest and StatsResponse proto definitions.

a-veitch:


grpc/instrumentation/v1alpha/monitoring.proto
```diff
@@ -103,12 +101,46 @@ message CanonicalRpcStats {
...
+  // descriptors will be sent back in the response. If no names are provided,
+  // then all Views present in the client system will be included in every
+  // response. If view_names is also provided, then Views matching the
+  // intersection of the two are returned.
+  // TODO(aveitch): Consider making this a list of regexes or prefix matches in
+  // the future.
+  repeated string measurement_names = 2;
+
+  // By default, the MeasurementDescriptors and ViewDescriptors corresponding to
+  // the Views that can returned in a StatsResponse will be included in the
```
ericgribkoff: s/can/are/ (or "can be"?)

grpc/instrumentation/v1alpha/monitoring.proto
```diff
@@ -103,12 +101,46 @@ message CanonicalRpcStats {
...
+  // first such response. Set this to true to have them not sent.
+  bool dont_include_descriptors_in_first_response = 3;
 }
 
+// This message contains all information relevant to a single View. It is used
+// in StatsResponse.
```
ericgribkoff: This is `StatsResponse`, so it doesn't make sense to says it's used in `StatsResponse`. What happened to the `ViewResponse` type that was wrapped in `StatsResponse` as a repeated field?

This is exactly the same as `CanonicalRpcStats.View` type. We should get rid of `CanonicalRpcStats.View` and just reuse `ViewResponse` (or `StatsResponse`). That can be done here or, if you prefer, I can submit it as a follow-up PR.

grpc/instrumentation/v1alpha/monitoring.proto
a-veitch: Done

grpc/instrumentation/v1alpha/monitoring.proto
a-veitch: We decided to have the 1:1 correspondance between the View data and the ViewDescriptor.

I've gone and removed the CanonicalRpcStats.View type.

a-veitch:
PTAL.

PR #4, opened 2017-03-13
BUILD file to make grpc-proto usable

makdharma:
Add a BUILD file so the monitoring.proto can be used in other
repositories that use bazel

### DIFF
--- a/grpc/instrumentation/v1alpha/BUILD
+++ b/grpc/instrumentation/v1alpha/BUILD
@@ -1,0 +1,44 @@
+# Copyright 2017, Google Inc.
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
+# met:
+#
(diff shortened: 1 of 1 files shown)


ericgribkoff:
A couple questions:

- This won't build as-is without a workspace file, correct?
- Where does `@submodule_grpc#bazel:grpc_build_system.bzl` come from?

I'm not sure it make sense to have a BUILD file here that can't actually build on its own. If this does actually build as-is, let me know how - I tried `bazel build` and that didn't work, but don't have a lot of experience with bazel.

ejona86:
grpc_proto_library is also a grpc/grpc-ism. That means it can't work in any repository other than grpc/grpc. Can the grpc/grpc build system not handle things if just a proto_library is defined here?

carl-mastrangelo:
I have since added a BUILD file, which can be used in the future.  It looks like this proto may be defunct, so I am closing this PR.

PR #12, opened 2018-05-02
make ALTS protos available in grpc-proto

carl-mastrangelo:
This moves the ALTS protos to a standard, common location.  In the future, when the golden files are updated inside of google, they will propagate to here first, and then down to each of the repos.  This should make it easier to keep all of them in sync.

cc: @cesarghali @jiangtaoli2016 @dfawley

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -33,3 +33,13 @@ proto_library(
     srcs = ["grpc/health/v1/health.proto"],
     visibility = ["//visibility:public"],
 )
+
+proto_library(
+    name = "alts_handshaker_proto",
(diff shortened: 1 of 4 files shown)


jiangtaoli2016:
@yihuazhang  FYI

jiangtaoli2016:
If there are changes here, how to propagate to each repo (grpc-core, grpc-java, grpc-go)? Manually?

carl-mastrangelo:
@jiangtaoli2016 manually for now, but I have a plan for that too.  Just wanted to make the source of truth this one for now, until we can make it automated.

PR #19, opened 2018-05-04
binarylog.proto: clarify semantics about half-closure and call end

zpencer:
This refines the proto to capture information about half closure and cancellations.

### DIFF
--- a/grpc/binlog/v1alpha/binarylog.proto
+++ b/grpc/binlog/v1alpha/binarylog.proto
@@ -28,6 +28,9 @@ message GrpcLogEntry {
   enum Type {
     UNKNOWN_TYPE = 0;
     SEND_INITIAL_METADATA = 1;
+    // On server side, a SERVER_CANCEL can still happen after sending
+    // trailers.
+    // On client side, this is means the client has half closed.
     SEND_TRAILING_METADATA = 2;
     SEND_MESSAGE = 3;
     RECV_INITIAL_METADATA = 4;
--- a/grpc/binlog/v1alpha/binarylog.proto
+++ b/grpc/binlog/v1alpha/binarylog.proto
@@ -37,7 +37,7 @@ message GrpcLogEntry {
     RECV_MESSAGE = 6;
     // Client side only.
     CLIENT_SEND_CANCEL = 7;
-    // Server side only. This is the last event for server side.
+    // Server side only. This means the RPC has now ended.
     SERVER_END_CALL = 8;
   };
 
--- a/grpc/binlog/v1alpha/binarylog.proto
+++ b/grpc/binlog/v1alpha/binarylog.proto
@@ -31,8 +31,14 @@ message GrpcLogEntry {
     SEND_TRAILING_METADATA = 2;
     SEND_MESSAGE = 3;
     RECV_INITIAL_METADATA = 4;
+    // On server side, this event means the client has half-closed.
+    // On client side, this is means the RPC has now ended.
     RECV_TRAILING_METADATA = 5;
     RECV_MESSAGE = 6;
+    // Client side only.
+    CLIENT_SEND_CANCEL = 7;
+    // Server side only. This is the last event for server side.
+    SERVER_END_CALL = 8;
   };
 
   // Enumerates the entity that generates the log entry
@@ -40,6 +46,17 @@ message GrpcLogEntry {
     UNKNOWN_LOGGER = 0;
     CLIENT = 1;
     SERVER = 2;
+  }
+
+  // On server side, SEND_TRAILING_METADATA does not end the call
+  // because the client can still cancel the call.  This enum shows
+  // the way the call actually ended.
+  enum ServerCallEnd {
+    UNKNOWN_END = 0;
+    // The call ended normally
+    NORMAL = 1;
+    // The call ended abnormally
+    ABNORMAL = 2;
   }
 
   Type type = 1;      // One of the above Type enum
@@ -64,6 +81,9 @@ message GrpcLogEntry {
 
   // true if payload does not represent the full message or metadata.
   bool truncated = 7;
+
+  // For SERVER_END_CALL only.
+  ServerCallEnd server_end = 8;
 };
 
 // Message payload, used by REQUEST and RESPONSE
--- a/grpc/binlog/v1alpha/binarylog.proto
+++ b/grpc/binlog/v1alpha/binarylog.proto
@@ -38,12 +38,8 @@ message GrpcLogEntry {
     // On client side, this is means the RPC has now ended.
     RECV_TRAILING_METADATA = 5;
     RECV_MESSAGE = 6;
-    // Client side only.
-    CLIENT_SEND_CANCEL = 7;
-    // Server side only. The RPC was cancalled, which may happen when
-    // a RST stream is received (client cancellation), a server side
-    // deadline is exceeded, an IO error occurred, etc.
-    SERVER_CANCEL = 8;
+    // The RPC is cancelled.
+    CANCEL = 7;
     // Server side only. This means the RPC has now ended. On server
     // side, SEND_TRAILING_METADATA is not necessarily the end of the
     // RPC, because a SERVER_CANCEL can still happen. A SERVER_END
--- a/grpc/binlog/v1alpha/binarylog.proto
+++ b/grpc/binlog/v1alpha/binarylog.proto
@@ -37,8 +37,16 @@ message GrpcLogEntry {
     RECV_MESSAGE = 6;
     // Client side only.
     CLIENT_SEND_CANCEL = 7;
-    // Server side only. This means the RPC has now ended.
-    SERVER_END_CALL = 8;
+    // Server side only. The RPC was cancalled, which may happen when
+    // a RST stream is received (client cancellation), a server side
+    // deadline is exceeded, an IO error occurred, etc.
+    SERVER_CANCEL = 8;
+    // Server side only. This means the RPC has now ended. On server
+    // side, SEND_TRAILING_METADATA is not necessarily the end of the
+    // RPC, because a SERVER_CANCEL can still happen. A SERVER_END
+    // marks the end of the RPC. This is not needed for client side
+    // because RECV_TRAILING_METADATA is always the last entry.
+    SERVER_END = 9;
   };
 
   // Enumerates the entity that generates the log entry
@@ -46,17 +54,6 @@ message GrpcLogEntry {
     UNKNOWN_LOGGER = 0;
     CLIENT = 1;
     SERVER = 2;
-  }
-
-  // On server side, SEND_TRAILING_METADATA does not end the call
-  // because the client can still cancel the call.  This enum shows
-  // the way the call actually ended.
-  enum ServerCallEnd {
-    UNKNOWN_END = 0;
-    // The call ended normally
-    NORMAL = 1;
-    // The call ended abnormally
-    ABNORMAL = 2;
   }
 
   Type type = 1;      // One of the above Type enum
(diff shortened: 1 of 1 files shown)


grpc/binlog/v1alpha/binarylog.proto
```diff
@@ -31,8 +31,14 @@ message GrpcLogEntry {
     SEND_TRAILING_METADATA = 2;
     SEND_MESSAGE = 3;
     RECV_INITIAL_METADATA = 4;
+    // On server side, this event means the client has half-closed.
+    // On client side, this is means the RPC has now ended.
     RECV_TRAILING_METADATA = 5;
     RECV_MESSAGE = 6;
+    // Client side only.
+    CLIENT_SEND_CANCEL = 7;
```
markdroth: I assume that this corresponds to an HTTP/2 RSET_STREAM?

grpc/binlog/v1alpha/binarylog.proto
```diff
@@ -31,8 +31,14 @@ message GrpcLogEntry {
...
+    // Server side only. This means the RPC has now ended.
+    SERVER_END_CALL = 8;
```
markdroth: It looks like this is conflating two cases:

1. The server receives a RSET_STREAM from the client after it has sent trailing metadata.  In this case, we probably don't need to log the event in the first place, because the server should consider the call complete once it sends trailing metadata.

2. The server receives a RSET_STREAM from the client before it has sent trailing metadata.  In this case, we could record the event as SERVER_RECV_CANCEL.

I think this representation would be simpler than SERVER_END_CALL, where you need to look at a separate value to understand what happened.

grpc/binlog/v1alpha/binarylog.proto
markdroth: What happens if the call is cancelled from the server side?  Would it make sense to split this up into SEND_CANCEL and RECV_CANCEL, either of which can be used on the client or server side?

zpencer: Correct, this corresponds to the client application requesting an explicit cancellation. But because binary logging happens at a high level, I think we should treat the RST_STREAM as an implementation detail.

grpc/binlog/v1alpha/binarylog.proto
zpencer: @ejona86 can you give some guidance here since you've been looking into the gRPC call semantics

grpc/binlog/v1alpha/binarylog.proto
zpencer: On client side, inspecting the status would be enough to determine if the server cancelled the call because an RPC always ends with some sort of status (even if locally generated by gRPC).

On server side, applications do not know why a RPC abnormally terminated. It can be that the client sent a cancellation, or something else like deframing failed, a ping timed out, or server shutdown was initiated. These are all conflated together as [onCancel](https://grpc.io/grpc-java/javadoc/io/grpc/ForwardingServerCallListener.html#onCancel--).

grpc/binlog/v1alpha/binarylog.proto
ejona86: For (1) I would caution you to realize there's the _service_ and the _grpc implementation_. The _service_ may have completed with trailers (which is SEND_TRAILING_METADATA), but the RST_STREAM (or I/O failure, etc) may arrive before that is sent on the wire. Note this is easy to trigger, because trailers are limited by flow control and thus are client-controlled on how long they take to send.

For (1), if the grpc implementation has actually sent the trailers then there would have been a ServerCallEnded{NORMAL}. Then receiving a RST_STREAM does nothing and wouldn't be logged at all, because the call is already complete.

The "recv" of SERVER_RECV_CANCEL isn't quite right, as the server may have triggered it itself, as is the case with Deadlines. Even an I/O failure seems a bit weird to call it SERVER_RECV_CANCEL. Now simply SERVER_CANCEL could be okay, but that's basically the same as ServerCallEnded{ABNORMAL}. The question around ServerCallEnded really seems to be whether ServerCallEnded{NORMAL} is a "thing."

zpencer: I'm looking at this from the context of the analyzer tooling, and there is an existing feature that goes through a log and reports on the current in flight RPCs. Having an explicit ServerCallEnded{NORMAL} sentinel would make that sort of thing much easier.

yashykt: Since we aren't thinking of using the same tools any longer, I don't see any use for replicating features.

I agree that it might be useful to know how the call ended - whether it was cancelled or something else. For this, I suggest that we log the status (the status code and string). Logging the status would be an indication that the call has been completed (doesn't matter whether it completed properly or cancelled or some other error). This also covers the case of SEND_CANCEL, RECV_CANCEL, etc. and also follows the idea that the filter will be treated as something closer to the application and everything it does is something that the users could have done themselves.

grpc/binlog/v1alpha/binarylog.proto
yashykt: Again, with directly logging the status, we would not need to log CLIENT_SEND_CANCEL.

grpc/binlog/v1alpha/binarylog.proto
zpencer: Logging the status that the server app sends is sufficient for normal terminations, but if the client cancels the RPC then no status or trailing metadata will be sent. I'm thinking of something like this:
1. client sends initial metaata
1. server receives client initial metadata
1. sever sends initial metadata
1. client sends cancel (RST_STREAM)
1. server receives cancel

The RPC is now done but no status was sent by the server application. Perhaps the C API uses status to express this? For Java it seems misleading to record that the call ended with a Status.CANCELLED.

Another case is the server can send out a Status.CANCELLED, but because it is a normal termination, the RPC should be recorded as ServerCallEnded{NORMAL}.

RE: tools
I agree we don't need to replicate features exactly but being able to report on the liveness of an RPC  seems like a basic use case.

yashykt: @zpencer Why would it be misleading? The RPC was cancelled. As to why it was cancelled, the details could be found in the message. For a cancel originating from the client side, it would be something like - Received RST_STREAM
As for the second case, why do we need to call that 'normal' termination? I would imagine the user to expect normal to be GRPC_STATUS_OK. For the case that you describe it should suffice to provide Cancelled status with the message that the user provided.

I am not against logging SEND_CANCEL which can be from either the server or the client, but it just seems redundant information. SEND_CANCEL does not seem to be adding anything new IMO.

The other alternative would be to have a SEND_CANCEL type and then a type to log the status code on the client CLIENT_STATUS_CODE, while on the server, it would just be a boolean of -
- true if the server perceives the call as terminated with Status::OK 
- false if it is anything else (includes server cancellation, client cancellation and all other errors)
But, in this approach, we are losing data on why it was not an OK status on the server

yashykt: As for the liveness of the RPC, the final status on the RPC would denote that the RPC has ended.

zpencer: For the first example: Are you suggesting that the server should log a SEND_TRAILING_METADATA message when nothing was sent? Based on the liveness comment it sounds to be the case. No status or metadata was sent, which is why it's misleading. I see two options: 1) lie and claim there was a SEND_TRAILING_METADATA when in fact there wasn't anything sent, or 2) we have some other type of event like ServerCallEnded{ABNORMAL}.

For the second case: normal vs abnormal refers to whether the call terminated by the server sending a Status+Metadata. So I wouldn't say it's redundant information.

yashykt: I am not suggesting claiming that there was a send_trailing_metadata. I am proposing separating the notion of status code at the server side from what was actually sent to indicating how it came to an end. This means that the status code observed at the server side and that observed at the client side can potentially be different for error cases. This would not mean that the server actually sent a trailing metadata with that status code. It's just how the RPC ended according to the server. Note that for the cases where the server actually sends the status in trailing metadata which is the normal behavior in your proposal, the status code would match.

Also, since we have a separate type for metadata which we log, it would be unambiguous whether metadata was sent or not. We just need to know when and how the call was terminated even if trailing metadata was not sent.

zpencer: Java has a similar sort of "server side close status" today, except it is an internal mechanism. Does the C API expose this status to applications?

zpencer:
There are 3 points discussed on this PR, so allow me to summarize where we are at today:

client side half closure:
Use SEND_TRAILING_METADATA to express this. I think we are in agreement. 

cancellations on client side:
I still think we should have a separate message for CLIENT_SEND_CANCEL even though we can inspect the RECV_TRAILING_METADATA message's status. Initiating a cancellation is a first class gRPC concept, and we should not need to interpret a Status.CANCELLED message to figure out whether the client application initiated it.

cancellations on server side:
A cancellation can arrive after sending status, and this is visible in all 3 languages' APIs AFAICT. Exposing the "internal server side status code" feels like a big API change.

Would we be more amenable to the idea below:
- keep CLIENT_SEND_CANCEL to mean client initiated cancellation
- add SERVER_CANCEL to mean represent all cancellations (client initiated + server generated)
- add a SERVER_END to mean that this RPC has now completely ended. 

A few examples of valid sequences:
server side happy case: ... SEND_TRAILING_METADATA SERVER_END
server side cancel: ... SERVER_CANCEL SERVER_END
server side cancel racing with status+trailers: ... SEND_TRAILING_METADATA SERVER_CANCEL SERVER_END

client side cancel: ... CLIENT_SEND_CANCEL [RECV_TRAILING_METADATA where status is Status.CANCELLED]

markdroth:
I am okay with CLIENT_SEND_CANCEL.  While we can capture this with SEND_TRAILING_METADATA, I agree that it would be useful to clearly indicate the difference between a normal half-close and a cancellation.

I do not think server-side cancellation needs a special enum value.  That can be captured by SEND_TRAILING_METADATA on the server with a status of CANCELLED.  After all, the client cannot tell the difference between the server cancelling and simply sending trailing metadata with status CANCELLED.

I continue to believe that SERVER_END_CALL is unnecessary and confusing.  I think we can capture what we need in a less confusing way with SERVER_RECV_CANCEL.

zpencer:
Updated the proto to have a CLIENT_SEND_CANCEL, SERVER_CANCEL (to encompass all cancellation types), and SERVER_END. I don't think there's any way around the SERVER_END message if we want to have an unambiguous terminal state for server side. This strictness will benefit consumers of the log who aren't using tools with any "end of rpc" heuristics. The asymmetry wrt client side is unfortunate. PTAL @markdroth

zpencer:
@markdroth can I get a LGTM on this, if there are no more outstanding issues. I think keeping the `SERVER_END` sentinel is valuable in making the end of stream unambiguous.

zpencer:
@markdroth ping

zpencer:
Had a chat with @ejona86 and the conclusion is that we don't really care so much about who initiated the cancellation.

This server side example shows how it's difficult to define who cancelled an RPC:

`client --a--> proxy --b--> server`

If the server sees a cancellation, it can be because:
- `client` explicitly sent a cancel
- connection `a` failed, and the proxy must propagate the failure via an explicit cancellation
- connection `b` failed, and the server locally generated a cancellation

So the proposal for the binlog proto:
- introduce a CANCEL message. On both client and server side it means that an RPC is cancelled.
- introduce a SERVER_END_MARKER, because on server side a CANCEL can still happen after sending trailers. A sentinel message gives a clear end to the message series.

Btw an analogous situation occurs on client side:

`client <--a-- proxy <--b-- server`

If a client sees a non OK status, it can be because:
- `server` explicitly sent it
- connection `b` failed, and the proxy must propgate the failure via an explicit status
- connection `a` failed, and the client must locally generate a status

zpencer:
Let's continue this discussion on https://github.com/grpc/grpc-proto/pull/29, where the existing enums will be revamped.

zpencer:
Replaced by https://github.com/grpc/grpc-proto/blob/master/grpc/binlog/v1/binarylog.proto

PR #20, opened 2018-05-04
binarylog: move required info from metadata to top level fields

zpencer:
Different languages may show different information to applications via
the metadata. Document this behavior and ensure that universally
required info is denormalized into the proto.

### DIFF
--- a/grpc/binlog/v1alpha/binarylog.proto
+++ b/grpc/binlog/v1alpha/binarylog.proto
@@ -52,7 +52,13 @@ message GrpcLogEntry {
   // The logger uses one of the following fields to record the payload,
   // according to the type of the log entry.
   oneof payload {
-    // Used by CLIENT_INIT_METADATA, SERVER_INIT_METADATA and TRAILING_METADATA
+    // Used by CLIENT_INIT_METADATA, SERVER_INIT_METADATA and
+    // TRAILING_METADATA. The contents may be different depending on
+    // the gRPC implementation. For example, C shows the http/2
+    // headers such as :method to applications but Java hides http/2
+    // implementation details.  Information that is guaranteed to be
+    // present for all gRPC implemetnations (such as the full method
+    // name) is always denormalized into GrpcLogEntry.
     Metadata metadata = 4;
     // Used by REQUEST and RESPONSE
     Message message = 5;
@@ -64,6 +70,9 @@ message GrpcLogEntry {
 
   // true if payload does not represent the full message or metadata.
   bool truncated = 7;
+
+  // The method name.
+  string full_method_name = 8;
 };
 
 // Message payload, used by REQUEST and RESPONSE
--- a/grpc/binlog/v1alpha/binarylog.proto
+++ b/grpc/binlog/v1alpha/binarylog.proto
@@ -53,12 +53,10 @@ message GrpcLogEntry {
   // according to the type of the log entry.
   oneof payload {
     // Used by CLIENT_INIT_METADATA, SERVER_INIT_METADATA and
-    // TRAILING_METADATA. The contents may be different depending on
-    // the gRPC implementation. For example, C shows the http/2
-    // headers such as :method to applications but Java hides http/2
-    // implementation details.  Information that is guaranteed to be
-    // present for all gRPC implemetnations (such as the full method
-    // name) is always denormalized into GrpcLogEntry.
+    // TRAILING_METADATA. The contents may be slightly different
+    // depending on the transport implementation. For example, some
+    // http2 based implementations may include extra information here
+    // like :path.
     Metadata metadata = 4;
     // Used by REQUEST and RESPONSE
     Message message = 5;
--- a/grpc/binlog/v1alpha/binarylog.proto
+++ b/grpc/binlog/v1alpha/binarylog.proto
@@ -78,8 +78,8 @@ message GrpcLogEntry {
   // encoding.
   string status_message = 10;
 
-  // the deadline in nanoseconds
-  uint64 deadline_ns = 11;
+  // the timeout in nanoseconds
+  uint64 timeout_ns = 11;
 
   // more things may go here like: authority, agent
 };
--- a/grpc/binlog/v1alpha/binarylog.proto
+++ b/grpc/binlog/v1alpha/binarylog.proto
@@ -18,6 +18,8 @@ syntax = "proto3";
 
 package grpc.binarylog.v1alpha;
 
+import "google/protobuf/duration.proto";
+
 option java_multiple_files = true;
 option java_package = "io.grpc.binarylog";
 option java_outer_classname = "BinaryLogProto";
@@ -67,7 +69,9 @@ message GrpcLogEntry {
   // true if payload does not represent the full message or metadata.
   bool truncated = 7;
 
-  // The method name. Always logged for every entry.
+  // The method name. Logged for the first entry:
+  // RECV_INITIAL_METADATA for server side or
+  // SEND_INITIAL_METADATA for client side.
   string method_name = 8;
 
   // status_code and status_message:
@@ -79,9 +83,11 @@ message GrpcLogEntry {
   string status_message = 10;
 
   // the timeout in nanoseconds
-  uint64 timeout_ns = 11;
+  google.protobuf.Duration timeout_ns = 11;
 
-  // more things may go here like: authority, agent
+  // The entry sequence id for this call. The first GrpcLogEntry has
+  // a value of 1, to disambiguate from an unset value.
+  uint32 sequence_id_within_call = 12;
 };
 
 // Message payload, used by REQUEST and RESPONSE
@@ -102,7 +108,17 @@ message Message {
 // A list of metadata pairs, used in the payload of CLIENT_INIT_METADATA,
 // SERVER_INIT_METADATA and TRAILING_METADATA
 // Implementations may omit some entries to honor the header limits
-// of GRPC_BINARY_LOG_CONFIG. Implementations can choose which entries are logged.
+// of GRPC_BINARY_LOG_CONFIG.
+// 
+// Implementations will log 'grpc-status-details-bin' unless truncated.
+//
+// Implementations will not log the following entries, and excluding them
+// will not count as a truncation:
+// - entries handled by grpc and are not user visible, such as those
+//   that begin with 'grpc-' or keys like 'lb-token'
+// - transport specific entries, including but not limited to:
+//   ':path', ':authority', 'content-encoding', 'user-agent', 'te', etc
+// - entries added for call credentials
 message Metadata {
   repeated MetadataEntry entry = 1;
 }
--- a/grpc/binlog/v1alpha/binarylog.proto
+++ b/grpc/binlog/v1alpha/binarylog.proto
@@ -18,6 +18,7 @@ syntax = "proto3";
 
 package grpc.binarylog.v1alpha;
 
+import "google/protobuf/any.proto";
 import "google/protobuf/duration.proto";
 
 option java_multiple_files = true;
@@ -78,18 +79,23 @@ message GrpcLogEntry {
   // Only present for SEND_TRAILING_METADATA on server side or
   // RECV_TRAILING_METADATA on client side.
   uint32 status_code = 9;
+
+  // The value of the 'grpc-status-details-bin' metadata key. If
+  // present, this is always an encoded 'google.rpc.Status' message.
+   google.protobuf.Any status_details = 10;
+
   // An original status message before any transport specific
   // encoding.
-  string status_message = 10;
+  string status_message = 11;
 
   // the RPC timeout
-  google.protobuf.Duration timeout = 11;
+  google.protobuf.Duration timeout = 12;
 
   // The entry sequence id for this call. The first GrpcLogEntry has a
   // value of 1, to disambiguate from an unset value. The purpose of
   // this field is to detect missing entries in environments where
   // durability or ordering is not guaranteed.
-  uint32 sequence_id_within_call = 12;
+  uint32 sequence_id_within_call = 13;
 };
 
 // Message payload, used by REQUEST and RESPONSE
@@ -112,8 +118,6 @@ message Message {
 // Implementations may omit some entries to honor the header limits
 // of GRPC_BINARY_LOG_CONFIG.
 // 
-// Implementations will log 'grpc-status-details-bin' unless truncated.
-//
 // Implementations will not log the following entries, and this is
 // not to be treated as a truncation:
 // - entries handled by grpc that are not user visible, such as those
--- a/grpc/binlog/v1alpha/binarylog.proto
+++ b/grpc/binlog/v1alpha/binarylog.proto
@@ -112,8 +112,8 @@ message Message {
 // 
 // Implementations will log 'grpc-status-details-bin' unless truncated.
 //
-// Implementations will not log the following entries, and excluding them
-// will not count as a truncation:
+// Implementations will not log the following entries, and this is
+// not to be treated as a truncation:
 // - entries handled by grpc and are not user visible, such as those
 //   that begin with 'grpc-' or keys like 'lb-token'
 // - transport specific entries, including but not limited to:
--- a/grpc/binlog/v1alpha/binarylog.proto
+++ b/grpc/binlog/v1alpha/binarylog.proto
@@ -52,13 +52,11 @@ message GrpcLogEntry {
   // The logger uses one of the following fields to record the payload,
   // according to the type of the log entry.
   oneof payload {
-    // Used by CLIENT_INIT_METADATA, SERVER_INIT_METADATA and
-    // TRAILING_METADATA. The contents may be slightly different
-    // depending on the transport implementation. For example, some
-    // http2 based implementations may include extra information here
-    // like :path.
+    // Used by {SEND,RECV}_INITIAL_METADATA and
+    // {SEND,RECV}_TRAILING_METADATA. This contains only the metadata
+    // from the application.
     Metadata metadata = 4;
-    // Used by REQUEST and RESPONSE
+    // Used by {SEND,RECV}_MESSAGE
     Message message = 5;
   }
 
@@ -69,8 +67,8 @@ message GrpcLogEntry {
   // true if payload does not represent the full message or metadata.
   bool truncated = 7;
 
-  // The method name.
-  string full_method_name = 8;
+  // The method name. Always logged for every entry.
+  string method_name = 8;
 
   // status_code and status_message:
   // Only present for SEND_TRAILING_METADATA on server side or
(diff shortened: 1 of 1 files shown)


grpc/binlog/v1alpha/binarylog.proto
```diff
@@ -52,7 +52,13 @@ message GrpcLogEntry {
   // The logger uses one of the following fields to record the payload,
   // according to the type of the log entry.
   oneof payload {
-    // Used by CLIENT_INIT_METADATA, SERVER_INIT_METADATA and TRAILING_METADATA
+    // Used by CLIENT_INIT_METADATA, SERVER_INIT_METADATA and
```
markdroth: Why don't we just say that all implementations will log it as part of metadata?  We should not have the format be different across languages.

zpencer: In Java, things like ":path" do not appear in the metadata object shown to applications. I believe the motivation was to hide http/2 specific implementation details from the interceptors API. This info instead copied into the [io.grpc.MethodDescriptor](https://grpc.io/grpc-java/javadoc/io/grpc/MethodDescriptor.html#getFullMethodName--). One of the things a11r mentioned about binlog is that it should be high level and do what a user application could have done, which is why I am using public APIs for logging the data.

zpencer: To expand on this, this is especially true when we're talking about in-process transports or transport types we have internally. The status, method, and other special information that is always present may be special cased and not actually put into the metadata. So maybe "language differences" is a misnomer here.

zpencer:
"The same thing" could also be to agree to pull it out to a separate proto field. There's value in always logging the method and status despite any metadata limits in place.

grpc/binlog/v1alpha/binarylog.proto
markdroth: I'm fine with the idea that we don't want this encoded in the metadata because that's not appropriate for all transports.  But we should not do this differently by language.  If we are going to introduce new fields for this information, then those fields should be used by all languages.

zpencer: Updated the proto so that gRPC information is captured as top level fields, rather than relying an implementation that puts them as fields of the Metadata.

Metadata is still a place where things can diverge, because right now C is logging it after things like `:path` are added.  Do we want to filter out this duplicate info to be consistent? I think the place where go might log is high level and would omit those fields as well. @dfawley is that correct?

zpencer: +1 on that

markdroth: In that case, please remove the part of this comment that says that the values differ based on the transport implementation.

grpc/binlog/v1alpha/binarylog.proto
```diff
@@ -61,6 +65,17 @@ message GrpcLogEntry {
   // Peer address information, will only be recorded in SEND_INITIAL_METADATA
   // and RECV_INITIAL_METADATA entries.
   Peer peer = 6;
+
+  // The method name. Always logged for every entry.
+  string full_method_name = 8;
```
markdroth: Let's just call this `method_name`.  The `full_` prefix doesn't add any value, since gRPC always uses absolute method names.

grpc/binlog/v1alpha/binarylog.proto
```diff
@@ -61,6 +65,17 @@ message GrpcLogEntry {
...
+
+  // status_code and status_message:
+  // Only present for SEND_TRAILING_METADATA on server side or
+  // RECV_TRAILING_METADATA on client side.
+  uint32 status_code = 9;
+  // An original status message before any transport specific
+  // encoding.
+  string status_message = 10;
```
markdroth: Do we also want to log status details?

grpc/binlog/v1alpha/binarylog.proto
zpencer: Done

grpc/binlog/v1alpha/binarylog.proto
zpencer: Are there more details beyond status_code and status_message?

markdroth: Yes.  We can support arbitrary binary details, encoded as a `google.rpc.Status` proto, transmitted as the value of the `grpc-status-details-bin` metadata key.

zpencer: Although `grpc-status-details-bin` begins with the special `grpc-` prefix, it is something that is set at the application level via a utility. I think it's fine not to give it a top level field, and just treat it like a normal metadata entry because gRPC's public API does not know about it. But I do think that our tooling should make it easy to inspect the contents.

grpc/binlog/v1alpha/binarylog.proto
```diff
@@ -52,15 +52,28 @@ message GrpcLogEntry {
...
 
   // Peer address information, will only be recorded in SEND_INITIAL_METADATA
   // and RECV_INITIAL_METADATA entries.
   Peer peer = 6;
+
+  // The method name. Always logged for every entry.
+  string method_name = 8;
+
+  // status_code and status_message:
+  // Only present for SEND_TRAILING_METADATA on server side or
```
yashykt: This won't cover the case where the rpc is cancelled. In those cases, we would still want to log some status about how it ended.

zpencer: Agreed, we need to come up with how to express cancellations on this PR https://github.com/grpc/grpc-proto/pull/19

grpc/binlog/v1alpha/binarylog.proto
markdroth: @yang-g, in C++, don't we encode the error details in `grpc::Status` instead of returning it as metadata?

If so, it's really part of our API and the fact that we transmit it as metadata on the wire should be just as abstracted as the fact that we transmit the status code and message the same way.

yang-g: The error details does go into the Status object. But that is just for the C++ API level. The C API does not see the error details as special part of API like the error code and message. I agree it could be better to pluck it out and put it in a separate proto field, though I can see it is easier to just log it as part of metadata too.

grpc/binlog/v1alpha/binarylog.proto
yang-g: It is not clear to me what is meant by "added by grpc library itself" as there could be filters putting in metadata as needed and some of the information could be useful?

zpencer: I think if the filter is something that is user accessible then it's fair to log it. But if the filter is a part of gRPC's internals then it's fine to omit that data, since the binary log should be capturing what the user application is doing.

markdroth: Since the intent here is to make binary logging be close to the application level rather than close to the wire level, I think it's fine to not log anything that is removed before the metadata is returned to the application.

grpc/binlog/v1alpha/binarylog.proto
markdroth: The C core API doesn't really matter; I think what matters is the C++ API, since that's what we expose to users.  (Also, I think the fact that the C-core API returns the error details in a different way is something we should fix.)

zpencer: @ejona86 @markdroth @dfawley PTAL

I audited Java and go and C:
- golang: [http2_client.go](https://github.com/grpc/grpc-go/blob/master/transport/http2_client.go#L356)
- java: [callers of Metadata.Key.of()](https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/Metadata.java#L566)
- C: [gen_static_metadata.py](https://github.com/grpc/grpc/blob/master/tools/codegen/core/gen_static_metadata.py) 

binlog should exclude:
- anything that begins with "grpc-": these are set by gRPC and are not be needed. Exceptions: grpc-trace-bin and grpc-tags-bin are a gray area because it's not clear whether to treat census as an application, I can see these going either way. "grpc-status-details-bin" should be included.
- "lb-token": This token is handled by grpc. In fact, for java, this happens in the load balancer and is not visible to interceptors.
- {http2, speedy} transport specific headers: see list below
- call credentials

binlog should include:
- "grpc-status-details-bin": spoke with a11r and determined it's ok to log this as a metadata field, and make the tooling responsible for parsing it. The status code and message should already match the grpc status.
- stickiness service config key: since this is application controlled, and could very well be a pre-existing application key that is also used for load balancing. 
- "user-agent", ":authority" both are ambiguous but from past discussions it sounds like we want to log these.

Example of grpc- keys that should be excluded if visible:
"grpc-timeout" // log as top level proto field: timeout_ns
"grpc-status"  // log as top level proto field
"grpc-tags-bin"  // questionable
"grpc-trace-bin"  // questionable
"grpc-previous-rpc-attempts"
"grpc-retry-pushback-ms"
"grpc-encoding"
"grpc-accept-encoding"

Should be excluded if visible:
"lb-token": lb is lower level than what the application sees

http/2 metadata that should be excluded:
":path"  // log as top level proto field: method_name
":status"  // grpc status should be provided
":authority"
":method"
":scheme"
"content-encoding"
"accept-encoding"
"content-type"
"te"
"user-agent"
// spdy only
":host"
":version"
// gen_static_metadata.py http2 metadata
content-disposition
expires
... etc

ejona86: SGTM

zpencer: Btw please pay special consideration to excluding call credentials. These look easy to exclude for java and golang but I am not familiar enough with C to to tell.

zpencer: For the two ambiguous grpc- keys (grpc-trace-bin and grpc-tags-bin) I propose to exclude it, and have `grpc-status-details-bin` be the only exception to the rule.

grpc/binlog/v1alpha/binarylog.proto
```diff
@@ -52,15 +52,36 @@ message GrpcLogEntry {
...
   }
 
   // Peer address information, will only be recorded in SEND_INITIAL_METADATA
   // and RECV_INITIAL_METADATA entries.
   Peer peer = 6;
+
+  // true if payload does not represent the full message or metadata.
+  bool truncated = 7;
+
+  // The method name. Always logged for every entry.
```
markdroth: Why every entry?  Seems like this will waste a lot of space, since there will generally be at least 6 entries per call (and arbitrarily many for streaming).

grpc/binlog/v1alpha/binarylog.proto
```diff
@@ -52,15 +52,36 @@ message GrpcLogEntry {
...
+  // status_code and status_message:
+  // Only present for SEND_TRAILING_METADATA on server side or
+  // RECV_TRAILING_METADATA on client side.
+  uint32 status_code = 9;
+  // An original status message before any transport specific
+  // encoding.
+  string status_message = 10;
+
+  // the timeout in nanoseconds
+  uint64 timeout_ns = 11;
```
markdroth: How about using `google.protobuf.Duration` for this?

grpc/binlog/v1alpha/binarylog.proto
markdroth: In general, this looks right to me.

I think the rule of thumb here should be "include only what would need to be specified by the application to replay the call".  That implies:

- Include `grpc-status-details-bin`, since that's specified by the application
- Include stickiness key, since that's specified by the application
- Exclude census metadata (will be added automatically to the replayed RPC by census)
- Exclude `user-agent` and `:authority` (will be automatically added to the replayed RPC by gRPC)

We should definitely exclude call credentials, but I don't think we can do this with a blanket rule, because call creds could use any arbitrary metadata key.  It will be up to individual implementations to make sure these are excluded.

carl-mastrangelo: @markdroth Would it be worth adding some sort of "redacted" header to here?  Something that indicates that call creds were present, but have been removed?

markdroth: What would the use case be for that?

grpc/binlog/v1alpha/binarylog.proto
zpencer: This works

grpc/binlog/v1alpha/binarylog.proto
markdroth: Why would a log message get dropped?  And if it does, since the information for the call will be incomplete in the log anyway, why do we care if the remaining info can be decoded?

grpc/binlog/v1alpha/binarylog.proto
carl-mastrangelo: Diagnotistic.  If  call fails due to bad creds, looking back in the logs would make it seem like the creds were never specified.  It would be similar to the truncated bit.

markdroth: If a call fails due to bad creds, it doesn't really matter whether the creds are missing or incorrect; either way, you will need to do further debugging to see what's actually being sent.  I'm not sure it really adds much to indicate this in the binary info logs.

grpc/binlog/v1alpha/binarylog.proto
zpencer: The underlying writer implementation that we use is allowed to drop writes when experiencing memory pressure. I also notice that it does not guarantee consistent ordering across all CPUs. It would be nice to still offer some info in the degraded mode, but there's no requirement to do this. 

Let's only log for the initial message, and not put too much effort into incompletely logged RPCs. I'm not sure how much of a problem they actually are.

grpc/binlog/v1alpha/binarylog.proto
carl-mastrangelo: Not to belabor this, but can't that argument be made for *any* header then?

grpc/binlog/v1alpha/binarylog.proto
zpencer: Once https://github.com/grpc/grpc-proto/pull/19 is sorted out, the final entry for the rpc (whether it is a normal termination or a cancellation) should probably log a count of how many entries are expected for this RPC.

grpc/binlog/v1alpha/binarylog.proto
markdroth: Sure.  Which is why we also don't include an indication that any other header has been excluded from the logs.

zpencer: A recent CL has added a truncation flag for headers: https://github.com/grpc/grpc-proto/commit/22fb07542ca06f954bb6f384ffbb19f40be3232a , should this not be logged at all?

markdroth: No, I'm fine with that; it indicates that the log's payload data (whether it be metadata or headers) may not be complete.  I just don't see a need to indicate that each individual header we're excluding is being excluded.

zpencer:
Reviewers: This should be everything sans cancellation, PTAL

grpc/binlog/v1alpha/binarylog.proto
```diff
@@ -64,6 +68,26 @@ message GrpcLogEntry {
...
+  // status_code and status_message:
+  // Only present for SEND_TRAILING_METADATA on server side or
+  // RECV_TRAILING_METADATA on client side.
+  uint32 status_code = 9;
+  // An original status message before any transport specific
+  // encoding.
+  string status_message = 10;
+
+  // the timeout in nanoseconds
+  google.protobuf.Duration timeout_ns = 11;
```
markdroth: No need for the `_ns` suffix here, now that the type is `google.protobuf.Duration`.

grpc/binlog/v1alpha/binarylog.proto
```diff
@@ -64,6 +68,26 @@ message GrpcLogEntry {
...
+
+  // The entry sequence id for this call. The first GrpcLogEntry has
+  // a value of 1, to disambiguate from an unset value.
+  uint32 sequence_id_within_call = 12;
```
markdroth: The comment for this should try to make it clear that the purpose for this is to allow the log reader to determine when there are missing entries.

grpc/binlog/v1alpha/binarylog.proto
```diff
@@ -84,7 +108,17 @@ message Message {
...
 // SERVER_INIT_METADATA and TRAILING_METADATA
 // Implementations may omit some entries to honor the header limits
-// of GRPC_BINARY_LOG_CONFIG. Implementations can choose which entries are logged.
+// of GRPC_BINARY_LOG_CONFIG.
+// 
+// Implementations will log 'grpc-status-details-bin' unless truncated.
+//
+// Implementations will not log the following entries, and this is
+// not to be treated as a truncation:
+// - entries handled by grpc and are not user visible, such as those
```
markdroth: s/and/that/

dfawley:
cc @MakMukhi 

grpc-go removes `grpc-status-details-bin` from the metadata handed back to the application.  Also, the application does not receive the serialized `grpc-status-details-bin` at all, only the _deserialized_ `google.rpc.Status` message.  If we have to include this in the metadata in binary logs, we would have to re-serialize it or (better) change/extend our API to be able to retrieve the serialized details.  I would prefer to log it as a separate field, and _ideally_ in message form (i.e. not serialized, but I understand this could be problematic for C/Java), rather than having it be the lone exception to the rule of removing `grpc-` headers from metadata.

zpencer:
Any objections to making it an Any? There are tools that understand how to unpack the encoded message from Any, so this would play better with them than a `bytes`.

I still think the payload can remain as `bytes` because Any's documentation does not mention anything about truncation.

dfawley:
> Any objections to making it an Any? There are tools that understand how to unpack the encoded message from Any, so this would play better with them than a bytes.

An `Any` where the `type_url` field is `"google.rpc.Status"` (or whatever the appropriate encoding of that is) should be fine.

markdroth:
I think we probably need to clarify the semantics of the status details in the gRPC API across languages before we can figure out the right way to encode it in the binary logging proto.  @ejona86 will be driving this discussion internally.

zpencer:
I changed the status details field to a `bytes` because:
- golang would have the serialized form at one point or another
- "casting" a bytes field to a typed message is possible, but using an `Any` makes it harder
- it is questionable whether tools properly unpack nested `Any` fields

zpencer:
@dfawley @markdroth PTAL

grpc/binlog/v1alpha/binarylog.proto
```diff
@@ -64,6 +68,33 @@ message GrpcLogEntry {
...
+  // RECV_TRAILING_METADATA on client side.
+  uint32 status_code = 9;
+
+  // The value of the 'grpc-status-details-bin' metadata key. If
+  // present, this is always an encoded 'google.rpc.Status' message.
+  bytes status_details = 10;
+
+  // An original status message before any transport specific
+  // encoding.
+  string status_message = 11;
```
markdroth: Nit: status_message should probably come before status_details.

PR #21, opened 2018-05-09
binlog: always log peer info

zpencer:
We can not always rely on seeing the initial metadata proto before any messages: https://paste.googleplex.com/5223926032498688

### DIFF
--- a/grpc/binlog/v1alpha/binarylog.proto
+++ b/grpc/binlog/v1alpha/binarylog.proto
@@ -58,8 +58,7 @@ message GrpcLogEntry {
     Message message = 5;
   }
 
-  // Peer address information, will only be recorded in SEND_INITIAL_METADATA
-  // and RECV_INITIAL_METADATA entries.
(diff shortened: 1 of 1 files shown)


yashykt:
Even though the order can be jumbled up, the initial metadata will be seen and logged at some point. I don't imagine it being that difficult to search for it. That being said, I also had the opinion earlier that peer info being logged everytime will make things easier for the reader. If I remember correctly, the pushback was that we should not be logging data unnecessarily.

zpencer:
It's not just out of orderness, any write could be dropped so we can never really be sure that the entire RPC is reconstructed from start to end. This extra bit of info will allow some basic analysis even in the degraded case.

zpencer:
If there is an entry that looks odd, the peer info will tell where to look next for more debug info.

markdroth:
It's quite common for a single client to send many RPCs to the same server.  Given that, I don't see how this change would actually help deal with ordering issues.

PR #22, opened 2018-05-18
binarylog: add truncation flag to metadata

zpencer:
We should log whether the Metadata was truncated. For messages, we
already have the `length` field that serves this purpose.


yashykt:
hmm if we are doing this for metadata, we should be doing it for messages too

PR #23, opened 2018-05-19
binarylogs: split peer address and port into separate fields

zpencer:
Using sockaddr_in is leaking C implementation details into the proto.
Since we really only care about address and port, let's just store
exactly that. This makes it much easier for tooling and non-C gRPC
implementations to deal with Peer.

### DIFF
--- a/grpc/binlog/v1alpha/binarylog.proto
+++ b/grpc/binlog/v1alpha/binarylog.proto
@@ -105,8 +105,12 @@ message Peer {
     PEER_UNIX = 3;
   };
   PeerType peer_type = 1;
-  string peer = 2;  // value depends on peer_type
-  uint16 port = 3; // only for PEER_IPV4 and PEER_IPV6
+  // value depends on peer_type
+  string peer = 2;
+  // only for PEER_IPV4 and PEER_IPV6
+  uint16 ip_port = 3;
+  // only for PEER_IPV6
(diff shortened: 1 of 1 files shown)


grpc/binlog/v1alpha/binarylog.proto
```diff
@@ -97,12 +97,13 @@ message MetadataEntry {
 message Peer {
   enum PeerType {
     UNKNOWN_PEERTYPE = 0;
-    PEER_IPV4 = 1;  // peer is struct sockaddr_in
-    PEER_IPV6 = 2;  // peer is struct sockaddr_in6
-    PEER_UNIX = 3;  // peer is struct sockaddr_un
+    PEER_IPV4 = 1;  // peer is 4 byte ipv4 address
```
yashykt: That won't be human readable

zpencer: I'm all for making things more readable. What do you think about using the [canonical ipv6](https://tools.ietf.org/html/rfc5952#section-4) address format?

zpencer:
@yashykt PTAL. The ipv6 scope is only  applicable for link local addresses, I don't think it's a priority to support it today but it will exist here for completeness.

zpencer:
uint16 is not a valid proto field type.
use a new field for "address" to make it easier to migrate to this new proto.

PR #28, opened 2018-06-22
Update LB proto comment again

AspirinSJL:
I changed this in grpc repo before. I feel it's worth keeping because it does clarify the port thing.

### DIFF
--- a/grpc/lb/v1/load_balancer.proto
+++ b/grpc/lb/v1/load_balancer.proto
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// This file defines an the GRPCLB LoadBalancing protocol.
+// This file defines the GRPCLB LoadBalancing protocol.
 //
 // The canonical version of this proto can be found at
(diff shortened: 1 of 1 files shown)


AspirinSJL:
Amended with nit.

grpc/lb/v1/load_balancer.proto
```diff
@@ -45,8 +45,10 @@ message LoadBalanceRequest {
 }
 
 message InitialLoadBalanceRequest {
-  // Name of load balanced service (IE, service.googleapis.com). Its
-  // length should be less than 256 bytes.
+  // The name of the load balanced service (e.g., balancer.service.com). The max
+  // length of the name is 256 bytes.
```
carl-mastrangelo: 256 or 255?

AspirinSJL: I just checked with the LB folks.  We actually don't see apparent limitation to this length in our LB implementation. We definitely need to discuss it and make the requirement clear. Before any decision is made, let me keep the old one.

PR #29, opened 2018-07-12
binlog: event type should be SERVER/CLIENT agnostic

zpencer:
This way of expressing events makes the proto much easier to consume. This should be an easy change for existing binlog implementations.

A few concrete examples of pain points:
- when replaying traffic against a real server, we must check whether the logs are recorded on client side or server side, and as a result play back sent messages vs received messages.
- for printing the message payloads, the protobuf API is expressed in terms of method input types and output types : https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.descriptor#MethodDescriptor . So we must go through the extra step of checking client vs server before we can print a message.

### DIFF
--- a/grpc/binlog/v1alpha/binarylog.proto
+++ b/grpc/binlog/v1alpha/binarylog.proto
@@ -39,25 +39,25 @@ message GrpcLogEntry {
 
   // Enumerates the type of event
   enum EventType {
-    UNKNOWN_EVENT_TYPE = 0;
+    EVENT_TYPE_UNKNOWN = 0;
     // Headers sent from client to server
-    CLIENT_HEADER = 1;
+    EVENT_TYPE_CLIENT_HEADER = 1;
     // Headers sent from server to client
-    SERVER_HEADER = 2;
+    EVENT_TYPE_SERVER_HEADER = 2;
     // Message sent from client to server
-    CLIENT_MESSAGE = 3;
+    EVENT_TYPE_CLIENT_MESSAGE = 3;
     // message sent from server to client
-    SERVER_MESSAGE = 4;
+    EVENT_TYPE_SERVER_MESSAGE = 4;
     // Signal that client is done sending
-    CLIENT_HALF_CLOSE = 5;
+    EVENT_TYPE_CLIENT_HALF_CLOSE = 5;
     // Trailers sent from server to client.
     // On client side, this is the last network event.
-    // On server side, a CANCEL can still happen after this.
-    SERVER_TRAILER = 6;
+    // On server side, a EVENT_TYPE_CANCEL can still happen after this.
+    EVENT_TYPE_SERVER_TRAILER = 6;
     // Signal that the RPC should be cancelled
-    CANCEL = 7;
+    EVENT_TYPE_CANCEL = 7;
     // A special marker to indicate that the RPC is done.
-    END_MARKER = 8;
+    EVENT_TYPE_END_MARKER = 8;
   }
 
   // Enumerates the entity that generates the log entry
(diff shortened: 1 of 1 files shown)


carl-mastrangelo:
@dfawley cc

grpc/binlog/v1alpha/binarylog.proto
```diff
@@ -35,7 +35,31 @@ message GrpcLogEntry {
     RECV_INITIAL_METADATA = 4;
     RECV_TRAILING_METADATA = 5;
     RECV_MESSAGE = 6;
-  };
+  }
+
+  // Enumerates the type of logs
+  enum EventType {
+    UNKNOWN_TYPE = 0;
```
dfawley: This won't compile in C++ due to the lack of enum namespacing.

grpc/binlog/v1alpha/binarylog.proto
```diff
@@ -54,29 +78,25 @@ message GrpcLogEntry {
   // The logger uses one of the following fields to record the payload,
   // according to the type of the log entry.
   oneof payload {
```
dfawley: Option C is to encode the message type within this oneof, and leave out the extra enum:

```
oneof payload {
   Metadata request_header = 4;
   Metadata response_header = 5;
   Message request_message = 6;
   ...
   bool half_close = ...
   ...
}
```

I remember I brought this up in an earlier review on this proto definition, but I don't recall the response.  Does this make real-world usage worse?

markdroth:
I don't like the terms "request" and "response", because they're too unary-centric.  I prefer the existing terms that we've already been using.

Moreover, it's not clear to me why this PR actually makes anything easier.  We already have a `logger` field to know whether a given message is logged by the server or the client.  So why do we need this to differentiate?

ejona86:
> I don't like the terms "request" and "response", because they're too unary-centric.

How so? A stream can simply have multiple request and response messages. Note that I used the request/response terminology in grpc/grpc#15460 

> I prefer the existing terms that we've already been using.

The problem is that the existing names are relative events. The same action changes depending on if you were on client or server. Changing to non-relative names is a huge simplification in my mind. Are there other non-relative names that would work better?

> We already have a logger field to know whether a given message is logged by the server or the client. So why do we need this to differentiate?

The problem is that all consumers of the data have to have custom logic for server vs client recordings, when really, they largely don't care where it was recorded.

Consider a record/playback system. You could record on client-side or server-side. You can play-back on client-side or server-side. A client-side playback can either: require a client-side recording or have tons of ifs to convert a server-side recording to the equivalent of client-side. With these changes such a playback system wouldn't generally need to observe the client/server bit.

grpc/binlog/v1alpha/binarylog.proto
```diff
@@ -35,7 +35,31 @@ message GrpcLogEntry {
...
+    RESPONSE_MESSAGE = 4;
+    // Signal that client is done sending
+    HALF_CLOSE = 5;
+    // Trailers sent from server to client.
+    // On client side, this is the final entry for an RPC.
+    // On server side, a CANCEL can still happen after this.
+    STATUS_TRAILERS = 6;
+    // Signal that the RPC should be cancelled
+    CANCEL = 7;
+    // A special event that only applicable to server side.
```
ejona86: Have client-side generate this too. It's trival for the client logger to do and it makes the log completely consistent between client and server.

zpencer: We can be consistent in the sense that the end of RPC is always marked with the end marker. But of course, the two side can still see different things in the event of a failure (client sees: `...> STATUS_TRAILERS, END_MARKER` but server sees: `... STATUS_TRAILERS, CANCEL, END_MARKER`).

grpc/binlog/v1alpha/binarylog.proto
```diff
@@ -35,7 +35,31 @@ message GrpcLogEntry {
...
     RECV_TRAILING_METADATA = 5;
     RECV_MESSAGE = 6;
-  };
+  }
+
+  // Enumerates the type of logs
+  enum EventType {
+    UNKNOWN_EVENT_TYPE = 0;
+    // Headers sent from client to server
+    REQUEST_HEADERS = 1;
```
dfawley: This is related to my comments in grpc/grpc#15460: can we standardize on "header" and "trailer" instead of "headers" and "trailers"?  This avoids the awkwardness of things like "the client sends a headers"

zpencer: Ack, I think this would be an improvement

zpencer:
- s/headers/header/, s/trailers/trailer/
- use the end marker on client side to be consistent
- avoid using the terms request/response for less awkwardness in streaming

grpc/binlog/v1alpha/binarylog.proto
```diff
@@ -35,7 +35,30 @@ message GrpcLogEntry {
...
+    // message sent from server to client
+    EVENT_TYPE_SERVER_MESSAGE = 4;
+    // Signal that client is done sending
+    EVENT_TYPE_CLIENT_HALF_CLOSE = 5;
+    // Trailers sent from server to client.
+    // On client side, this is the last network event.
+    // On server side, a EVENT_TYPE_CANCEL can still happen after this.
+    EVENT_TYPE_SERVER_TRAILER = 6;
+    // Signal that the RPC should be cancelled
+    EVENT_TYPE_CANCEL = 7;
```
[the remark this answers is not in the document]
zpencer: I think we addressed this one offline: `EVENT_TYPE_CANCEL` is used to mean an application initiated cancellation on client side, and on server side we don't care whether it was explicitly initiated by the client app or if it is locally generated in response to a failure.

zpencer:
Replaced by https://github.com/grpc/grpc-proto/blob/master/grpc/binlog/v1/binarylog.proto

PR #30, opened 2018-07-13
Add worker_server and benchmark_service protos

kpayson64:


grpc/testing/benchmark_service.proto
```diff
@@ -0,0 +1,44 @@
...
+syntax = "proto3";
+
+import "grpc/testing/messages.proto";
+
+package grpc.testing;
+
+service BenchmarkService {
+  // One request followed by one response.
+  // The server returns the client payload as-is.
+  rpc UnaryCall(SimpleRequest) returns (SimpleResponse);
```
ejona86: It's pretty annoying that SimpleRequest/SimpleResponse aren't included in this file. It makes it less legible.

kpayson64: SimpleRequest/SimpleResponse are shared with the interop services, so they need their own proto file.

ejona86: I see. These proto files were weird to begin with, as all the messages used by ReportQpsScenarioService and WorkerService only used in that one service, but the messages exist in a different file. That's probably because this is a pretty old file, like test.proto, back before we had our infrastructure worked out to have messages in the same file.

Sight... report_qps_scenario_service.proto is such a useless file.

BUILD.bazel
```diff
@@ -16,6 +16,29 @@ package(default_visibility = [
...
+proto_library(
+    name = "stats_proto",
+    srcs = ["grpc/core/stats.proto"],
+)
+
+proto_library(
+    name = "benchmark_service_proto",
+    srcs = [
+        "grpc/testing/benchmark_service.proto",
+        "grpc/testing/messages.proto",
```
ejona86: Messages.proto should not be in srcs here. It should get its own proto_library, since it will be used by test.proto as well.

BUILD.bazel
```diff
@@ -16,6 +16,29 @@ package(default_visibility = [
...
+    ],
+)
+
+proto_library(
+    name = "worker_service_proto",
+    srcs = ["grpc/testing/worker_service.proto",
+            "grpc/testing/control.proto",
```
ejona86: This also doesn't look like it should be included, as it will be used by report_qps_scenario_service.proto.

BUILD.bazel
kpayson64: Done

BUILD.bazel
kpayson64: Done

PR #31, opened 2018-07-30
Fix binarylog's java_package to include version

ejona86:
Since this is a breaking change, it will need to be coordinated with the Java sync.

grpc/grpc-java#4704


ejona86:
@zpencer, this needed to be merged and managed at the same time as #4704. Can you please review #4704?

PR #33, opened 2018-08-07
Add Watch method to health checking service.

markdroth:


markdroth:
Added a comment about how to handle call termination.  PTAL.

carl-mastrangelo:
@markdroth can you import this change internally?  There should be a script for you to use.

PR #40, opened 2018-10-25
Add result cardinality limit to requests.

zpencer:
This requires tweaking how pagination works. Otherwise, it would not
be possible to paginate 1 result at a time. This implies that 0 MUST
NOT be a valid ID for any channelz entity.

See https://github.com/grpc/proposal/pull/114

### DIFF
--- a/grpc/channelz/v1/channelz.proto
+++ b/grpc/channelz/v1/channelz.proto
@@ -177,6 +177,7 @@ message SubchannelRef {
 
 // SocketRef is a reference to a Socket.
 message SocketRef {
+  // The globally unique id for this socket.  Must be a positive number.
   int64 socket_id = 3;
   // An optional name associated with the socket.
   string name = 4;
--- a/grpc/channelz/v1/channelz.proto
+++ b/grpc/channelz/v1/channelz.proto
@@ -435,9 +435,14 @@ service Channelz {
 }
 
 message GetTopChannelsRequest {
-  // start_channel_id indicates that only channels at or above this id should be
+  // start_channel_id indicates that only channels above this id should be
   // included in the results.
   int64 start_channel_id = 1;
+
+  // If non-zero, the server will return a page of results whose
+  // cardinality is equal to or less than this value. If zero, the
+  // server will choose a reasonable page size.
+  uint32 limit = 2;
 }
 
 message GetTopChannelsResponse {
@@ -451,9 +456,14 @@ message GetTopChannelsResponse {
 }
 
 message GetServersRequest {
-  // start_server_id indicates that only servers at or above this id should be
+  // start_server_id indicates that only servers above this id should be
   // included in the results.
   int64 start_server_id = 1;
+
+  // If non-zero, the server will return a page of results whose
+  // cardinality is equal to or less than this value. If zero, the
+  // server will choose a reasonable page size.
+  uint32 limit = 2;
 }
 
 message GetServersResponse {
@@ -468,9 +478,14 @@ message GetServersResponse {
 
 message GetServerSocketsRequest {
   int64 server_id = 1;
-  // start_socket_id indicates that only sockets at or above this id should be
+  // start_socket_id indicates that only sockets above this id should be
   // included in the results.
   int64 start_socket_id = 2;
+
+  // If non-zero, the server will return a page of results whose
+  // cardinality is equal to or less than this value. If zero, the
+  // server will choose a reasonable page size.
+  uint32 limit = 3;
 }
 
 message GetServerSocketsResponse {
--- a/grpc/channelz/v1/channelz.proto
+++ b/grpc/channelz/v1/channelz.proto
@@ -436,19 +436,23 @@ service Channelz {
 }
 
 message GetTopChannelsRequest {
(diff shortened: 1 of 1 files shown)


grpc/channelz/v1/channelz.proto
```diff
@@ -435,9 +436,14 @@ service Channelz {
 }
 
 message GetTopChannelsRequest {
```
lyuxuan: I don't think `GetTopChannelRequest` really need the limit, as we also have the `GetChannel` API.

grpc/channelz/v1/channelz.proto
```diff
@@ -435,9 +436,14 @@ service Channelz {
...
-  // start_channel_id indicates that only channels at or above this id should be
+  // start_channel_id indicates that only channels above this id should be
```
lyuxuan: why changing to above instead of at or above?

zpencer: This change is required to allow paginating 1 result at a time.
Let's suppose there are 3 channels: `[1,2,3]`. 
First call: `start_channel_id: 0, limit:1` --> First result: (channel 1)

The pagination token is generated on client side today (although I wouldn't be opposed to changing it to being generated by the server). So the client takes the last element from the result page and uses it as the pagination token:
Next call: `start_channel_id: 1, limit: 1` --> (channel 1)
We can never move past the first result.

If we change the definition, then the next result pages would contain (channel 2) and (channel 3) respectively.

carl-mastrangelo: I agree with @lyuxuan here, though getting server help in picking the next token would be useful.    The last item received is used (perhaps indirectly) to construct the next request.    The limit size should never modify the correctness of the approach, so I think the verbiage change here is incorrect.

grpc/channelz/v1/channelz.proto
```diff
@@ -435,9 +436,14 @@ service Channelz {
...
   // included in the results.
   int64 start_channel_id = 1;
+
+  // If non-zero, the server will return a page of results whose
+  // cardinality is equal to or less than this value. If zero, the
+  // server will choose a reasonable page size.
+  uint32 limit = 2;
```
carl-mastrangelo: Can we use int64?   We have used int64 in most other parts of this API.  The result size will likely never be that big.

grpc/channelz/v1/channelz.proto
```diff
@@ -435,9 +436,14 @@ service Channelz {
...
 
 message GetTopChannelsRequest {
-  // start_channel_id indicates that only channels at or above this id should be
+  // start_channel_id indicates that only channels above this id should be
   // included in the results.
   int64 start_channel_id = 1;
+
+  // If non-zero, the server will return a page of results whose
+  // cardinality is equal to or less than this value. If zero, the
+  // server will choose a reasonable page size.
```
carl-mastrangelo: Perhaps max_results?   limit is kinda ambiguous (is this a limit on the ids?).   Also, I think this needs to be specified for negative values.

grpc/channelz/v1/channelz.proto
```diff
@@ -435,9 +436,14 @@ service Channelz {
...
+  uint32 limit = 2;
 }
 
 message GetTopChannelsResponse {
```
carl-mastrangelo: not your change, but I think this doc needs to say that there must be at least one channel returned, or else end must be set.  (to avoid a possible infinite loop where the client requests channels, but the server returns none and doesn't set end.

grpc/channelz/v1/channelz.proto
```diff
@@ -435,9 +436,14 @@ service Channelz {
 }
 
 message GetTopChannelsRequest {
-  // start_channel_id indicates that only channels at or above this id should be
+  // start_channel_id indicates that only channels above this id should be
   // included in the results.
   int64 start_channel_id = 1;
+
+  // If non-zero, the server will return a page of results whose
+  // cardinality is equal to or less than this value. If zero, the
```
carl-mastrangelo: nit: cardinality is a big word that some none-english speakers may not understand.  Consider picking simpler, less apt words.  (Chrome even highlights it as misspelled for me).

grpc/channelz/v1/channelz.proto
zpencer: Per IRL discussion, let's document how the client must generate pagination tokens to make this work.

grpc/channelz/v1/channelz.proto
zpencer: Changed to be easier to understand.

grpc/channelz/v1/channelz.proto
```diff
@@ -437,12 +438,21 @@ service Channelz {
 message GetTopChannelsRequest {
   // start_channel_id indicates that only channels at or above this id should be
   // included in the results.
+  // To request the first page, this must be set to 0. To request
```
carl-mastrangelo: s/must/should/   Setting it to 1 should have identical results, since the bounds are inclusive

zpencer: You're right.

PR #41, opened 2018-10-26
Update interops test proto to match grpc-java.

creamsoup:
The test.proto is adapted from grpc-java, and stubby4 server for cloud to cloud 
interops test gets same update. 

- Using protobuf provided BoolValue instead of custom one. 
  the issue in the command is closed, and interops test passed
- This change is backward compatible.


ejona86:
I'd feel much better if we delayed making the wrappers change until after the various languages were all pulling from the same proto. We've had trouble with the well-known protos in a few languages, since our build infrastructure is... weird... in places.

ejona86:
My statement applies to Empty as well.

creamsoup:
changed back to use custom message types. I tested the previous version in go / c++ / python / ruby, etc. the well-known protos seem to work fine. but i agree that it still seems wiser to sync between languages first.

grpc/testing/messages.proto
```diff
@@ -31,6 +33,12 @@ message BoolValue {
 enum PayloadType {
   // Compressable text format.
   COMPRESSABLE = 0;
+
+  // Uncompressable binary format.
+  UNCOMPRESSABLE = 1;
```
ejona86: These were removed in C in https://github.com/grpc/grpc/pull/6839/files#r66543026

The DEPRECATED comments were only recently deleted in https://github.com/grpc/grpc/pull/15711 . We could finish the deprecation, since the enum is useless with only one entry.

creamsoup:
for the BoolValue, grpc-java actually uses well known protos. Since we want to keep the custom one due to the past issue, changing the java one to use custom bool value.

ejona86:
grpc-java using BoolValue for the moment sounds fine.

PR #42, opened 2018-10-26
Add 'summary' field to GetSocketRequest

zpencer:
If set, only return a high level summary.

I don't think the security info is relevant either for the high level view of a socket. I can revert this if we disagree.

Even though any field can be absent in proto3, it's probably still a good idea to document what 'summary' should suppress. Otherwise it's not clear if socket addresses, `last_local_stream_created_timestamp`, or `last_message_sent_timestamp` is expected to be present.

### DIFF
--- a/grpc/channelz/v1/channelz.proto
+++ b/grpc/channelz/v1/channelz.proto
@@ -234,9 +234,8 @@ message Socket {
   Address local = 3;
   // The remote bound address.  May be absent.
   Address remote = 4;
-  // Security details for this socket.  May be absent if not available, if
-  // there is no security on the socket, or if 'summary' is set on
-  // GetSocketRequest.
+  // Security details for this socket.  May be absent if not available, or
+  // there is no security on the socket.
   Security security = 5;
 
   // Optional, represents the name of the remote endpoint, if different than
--- a/grpc/channelz/v1/channelz.proto
+++ b/grpc/channelz/v1/channelz.proto
@@ -234,8 +234,9 @@ message Socket {
   Address local = 3;
   // The remote bound address.  May be absent.
   Address remote = 4;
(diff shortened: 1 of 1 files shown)


zpencer:
@lyuxuan @menghanl @ncteisen

grpc/channelz/v1/channelz.proto
```diff
@@ -234,8 +234,9 @@ message Socket {
   Address local = 3;
   // The remote bound address.  May be absent.
   Address remote = 4;
-  // Security details for this socket.  May be absent if not available, or
-  // there is no security on the socket.
+  // Security details for this socket.  May be absent if not available, if
+  // there is no security on the socket, or if 'summary' is set on
```
markdroth: I don't think we should add this note for any individual fields.  Let's just leave it entirely up to the implementation as to which fields are populated when `summary` is true.

carl-mastrangelo: I think implementer will look to this proto to see what fields they should include.  Protos are documentation.

markdroth: I think this is over-specifying the API, but if everyone else agrees, I won't make a big deal out of it.

However, if we are going to specify which fields are associated with the `summary` bit, then we need to agree on which ones should be in that category.  Up until now, we've only discussed the socket options, since it's expensive to get those from the kernel.  Why is security now being included?  What's expensive about that?

zpencer: The TLS certificates can be relatively large, but they are not expensive to access. If the goal of `summary` is to only return what is needed for the overview, then it may make sense to omit this. But on the other hand, this may be a TLS only problem because other types of `Security` info may not be so large (like ALTS), and we may want the client identity to be present in the overview.

PR #44, opened 2019-02-23
adds java_proto_library rules for all public protos.

perezd:
This makes it possible for grpc-java to properly depend on the java protos it needs to in its services/ source tree.

NOTE: This also upgrades protobuf to a newer version and fixes more recent Bazel deprecation errors.
Everything builds fine.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -70,10 +70,22 @@ proto_library(
     ],
 )
 
+java_proto_library(
+    name = "channelz_java_proto",
+    visibility = ["//visibility:public"],
+    deps = [":channelz_proto"],
+)
+
 proto_library(
     name = "health_proto",
     srcs = ["grpc/health/v1/health.proto"],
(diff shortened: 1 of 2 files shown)


ejona86:
@carl-mastrangelo, I thought we we were going to add java_proto_library and java_grpc_library targets to grpc-proto, similar to what we would do internally. So, no, I don't remember the reason.

perezd:
I didn't put grpc_java_library entries here because I was not sure if cyclic workspace dependencies are OK or not so for now I just put those here: https://github.com/grpc/grpc-java/pull/5384

perezd:
Friendly ping, anything blocking this?

carl-mastrangelo:
FYI, we haven't considered the stability of this in terms of versioning, so we may accidentally break it until then.

carl-mastrangelo:
@ejona86 could I get a second LGTM?

PR #45, opened 2019-03-04
import envoy proto files for xds

dapengzhang0:
Files imported by running `third_party/envoy/import.sh` and `third_party/protoc-gen-validate/import.sh`

### DIFF
--- a/third_party/envoy/LICENSE
+++ b/third_party/envoy/LICENSE
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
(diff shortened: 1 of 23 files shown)


ejona86:
Why are these being added?

dapengzhang0:
@dapengzhang0 because go wants to share the same import.

ejona86:
This is the canonical home for gRPC's protos. I'm worried about having other random protos included. That could be understandable if there was going to be a grpc-specific proto that used them, but as it is, it seems maybe just the import.sh script should live here (although C-core repo may also be a fine place).

ejona86:
It's also a bit weird to make a copy of the protos here to then just copy them again into our language-specific repos.

dfawley:
We should not be copying them into the language-specific repos if they are also here.  One argument in favor of this approach is to make sure all languages are using the same version of the protos.

lyuxuan:
Language-specific repos won't need a copy if we have them here here.

dapengzhang0:
I found that c-core repo has introduced submodule https://github.com/grpc/grpc/pull/17210. I not sure how the gogoproto issue is resovled in c-core,  + @vishalpowar . In google3, the envoy import script just removes all the gogoproto extensions.

ejona86:
> We should not be copying them into the language-specific repos if they are also here.

@dfawley, how would you use them from Go? There's no generated code here.

> One argument in favor of this approach is to make sure all languages are using the same version of the protos.

We won't all be using the same version of this repo, so I don't see how this applies. Following the source-of-truth, that is envoy, seems to be the important thing. And all users of the script would be doing that.

dfawley:
> @dfawley, how would you use them from Go? There's no generated code here.

We use all the .proto files in this location in the grpc-go repo.  We generate the code based on these protos and check in the .pb.go files into our repo.  We also have a nightly test to make sure they are kept up-to-date.

> We won't all be using the same version of this repo, so I don't see how this applies.

We ensure we are always using the latest version submitted here via testing.  Java and C could do the same.

> Following the source-of-truth, that is envoy, seems to be the important thing. And all users of the script would be doing that.

I would feel less comfortable having tests that fail when envoy updates their protos, which may happen more often than we are concerned with.

That said, I'm OK with us agreeing on a version of these proto files (e.g. as a git commit hash in their repo) to use in all 3 languages, and updating that agreed-upon version as necessary.  I agree that putting these files here isn't necessary, and I found it a bit unusual at first, too.

carl-mastrangelo:
Thus begins the monorepo.

dapengzhang0:
Made agreement in offline discussion. This is not needed. Go can import from envoy repo directly.

PR #55, opened 2019-06-11
Add optional grpclb_path field to SimpleResponse

apolcyn:
This is in order to add a couple of new tests around grpclb fallback

Test descriptions already discussed offline

### DIFF
--- a/grpc/testing/messages.proto
+++ b/grpc/testing/messages.proto
@@ -48,6 +48,21 @@ message EchoStatus {
   string message = 2;
 }
 
+// The type of route that a client took to reach a server w.r.t. gRPCLB.
+// The server must fill in "fallback" if it detects that the RPC reached
+// the server via the "gRPCLB fallback" path, and "backend" if it detects
+// that the RPC reached the server via "gRPCLB backend" path (i.e. if it got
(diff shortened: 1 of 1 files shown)


grpc/testing/messages.proto
```diff
@@ -94,6 +97,13 @@ message SimpleResponse {
   // Server ID. This must be unique among different server instances,
   // but the same across all RPC's made to a particular server instance.
   string server_id = 4;
+  // gRPCLB Path. This takes the value of either "fallback" or "backend".
+  // The server must fill in "fallback" if it detects that the RPC reached
+  // the server via the "gRPCLB fallback" path, and "backend" if it detects
+  // that the RPC reached the server via "gRPCLB backend" path (i.e. if it got
+  // the address of this server from the gRPCLB server BalanceLoad RPC). Exactly
+  // how this detection is done is server dependant.
+  string grpclb_path = 5;
```
zhangkun83: The field name is not self-explanatory and "path" leads me to think of something else. How about "route_type"? You may also consider using enums instead of strings.

apolcyn: Good point!

I've renamed from path to route_type and changed to an enum

PR #58, opened 2019-06-26
Service config proto

markdroth:
This is not actually used by anything externally, but we've had several people ask for us to publish this as a reference.

### DIFF
--- a/grpc/service_config/service_config.proto
+++ b/grpc/service_config/service_config.proto
@@ -228,16 +228,25 @@ message LoadBalancingConfig {
     // a round_robin policy to pick the backend within that cluster), its
     // configuration message may include a nested instance of the
     // LoadBalancingConfig message to configure the nested LB policy.
+
     PickFirstConfig pick_first = 4 [json_name = "pick_first"];
+
     RoundRobinConfig round_robin = 1 [json_name = "round_robin"];
-    // xds field should not be used unless people know what they're
-    // doing.
-    XdsConfig xds = 2;
+
+    // gRPC lookaside load balancing.
+    // This will eventually be deprecated by the new xDS-based local
+    // balancing policy.
     GrpcLbConfig grpclb = 3;
-    // xds_experimental field should only be used for testing the xds
-    // lbpolicy on the gRPC client.
-    // TODO(rekarthik): remove this field after xds migration is done.
+
+    // EXPERIMENTAL -- DO NOT USE
+    // xDS-based load balancing.
+    // The policy is known as xds_experimental while it is under development.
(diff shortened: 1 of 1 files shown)


grpc/service_config/service_config.proto
```diff
@@ -0,0 +1,315 @@
+// Copyright 2016 The gRPC Authors
```
dfawley: Is this the proper date?

grpc/service_config/service_config.proto
```diff
@@ -0,0 +1,315 @@
...
+    PickFirstConfig pick_first = 4 [json_name = "pick_first"];
+    RoundRobinConfig round_robin = 1 [json_name = "round_robin"];
+    // xds field should not be used unless people know what they're
+    // doing.
+    XdsConfig xds = 2;
+    GrpcLbConfig grpclb = 3;
+    // xds_experimental field should only be used for testing the xds
+    // lbpolicy on the gRPC client.
+    // TODO(rekarthik): remove this field after xds migration is done.
+    XdsConfig xds_experimental = 5 [json_name = "xds_experimental"];
```
dfawley: Should this be removed?

grpc/service_config/service_config.proto
```diff
@@ -0,0 +1,315 @@
...
+    // message for that LB policy may be empty.
+    //
+    // Note that if an LB policy contains another nested LB policy
+    // (e.g., a gslb policy picks the cluster and then delegates to
+    // a round_robin policy to pick the backend within that cluster), its
+    // configuration message may include a nested instance of the
+    // LoadBalancingConfig message to configure the nested LB policy.
+    PickFirstConfig pick_first = 4 [json_name = "pick_first"];
+    RoundRobinConfig round_robin = 1 [json_name = "round_robin"];
+    // xds field should not be used unless people know what they're
```
dfawley: Clean up this comment?

grpc/service_config/service_config.proto
markdroth: Yup.  That's when I created the current version of the proto internally.

grpc/service_config/service_config.proto
markdroth: Done.

grpc/service_config/service_config.proto
markdroth: I think it's fine to leave it here.  Given the name "experimental", anyone who tries to use it will not be expecting anything to work.

Note that once this change gets imported, I intend to replace our internal copy of this proto with this one, so the two need to be the same.  We are using the xds_experimental field during development.

PR #61, opened 2019-10-08
Add Travis CI support

ejona86:
Also fix the build for newer versions of Bazel, just so that travis-ci
can actually succeed.

tools/bazel is a copy of https://github.com/grpc/grpc/blob/master/tools/bazel


tools/bazel
```diff
@@ -0,0 +1,64 @@
...
+  "Darwin x86_64")
+    suffix=darwin-x86_64
+    ;;
+  *)
+    echo "Unsupported architecture: $(uname -sm)"
+    exit 1
+    ;;
+esac
+
+filename="bazel-$VERSION-$suffix"
```
dfawley: `...${VERSION}-${suffix}`.  Also, more braces below please.

tools/bazel
```diff
@@ -0,0 +1,64 @@
...
+
+# Keeping up with Bazel's breaking changes is currently difficult.
+# This script wraps calling bazel by downloading the currently
+# supported version, and then calling it. This way, we can make sure
+# that running bazel will always get meaningful results, at least
+# until Bazel 1.0 is released.
+# NOTE: This script relies on bazel's feature where //tools/bazel
+# script can be used to hijack "bazel" invocations in given workspace.
+
+set -e
```
dfawley: Consider: `set -x` for debugging purposes.

tools/bazel
ejona86: This script is verbatim from https://github.com/grpc/grpc/blob/master/tools/bazel . I'd rather keep it that way.

tools/bazel
ejona86: No, please, no.

tools/bazel
dfawley: :open_mouth:

```
.travis.yml:

script:
- bash <(curl https://raw.githubusercontent.com/grpc/grpc/master/tools/bazel)
```

?  Or pick a commit hash instead of `master`.

ejona86: It's placement in the repository is also magical. If you have bazel installed on your machine and run "bazel build" this script will choose the version of bazel to use, which may be different from what is installed on your machine.

If i'm scripting it in the travis.yml, I would just download the current bazel binary and not use the script at all.

I don't expect this to last. I personally like bazelisk. But this is _something_ that works and isn't NIH and can be iterated on.

PR #66, opened 2019-11-22
Added max_frame_size field for implementing frame size negotiation in ALTS

ashithasantosh:
Design doc - go/seal-frame-size-negotiation-alts


jiangtaoli2016:
@dfawley FYI. We will update https://github.com/grpc/grpc-go/tree/master/credentials/alts/internal/proto/grpc_gcp shortly.

PR #76, opened 2020-03-11
Add a new "targets" field for RouteLookupResponse 

cdmetcalf:


grpc/lookup/v1/rls_config.proto
```diff
@@ -212,4 +212,11 @@ message RouteLookupConfig {
     ASYNC_LOOKUP_DEFAULT_TARGET_ON_MISS = 3;
   }
   RequestProcessingStrategy request_processing_strategy = 10;
+
+  // Set to true to run the RLS client in "dark launch" mode, where it will do
+  // all the regular RLS processing (caching, doing RPC to lookup service,
+  // monitoring, logging, etc) but will *not* modify the actual target.
+  // Setting this flag enables a backend service to understand how load will
+  // shift once RLS is fully enabled.
+  bool enable_dark_launch = 11;
```
markdroth: Playing devil's advocate: Why is this necessary?  Can't a service owner accomplish the same thing by having the RLS server always return the same target but add a header indicating what target it would have chosen?

If there's really a compelling reason to do this, we can probably do it, but I would prefer to avoid adding more knobs to an already complicated design unless there's really a compelling advantage to doing so.

cdmetcalf: The distinction is that dark launch mode lets you attribute the actual amount of traffic by looking at the proxy monitoring for requests.  The RLS client caching will shield that information from the RLS server so it won't know how popular specific request keymaps are otherwise.   The proxy implementation is trivial (basically a one-line change) and  I'd hope it would be similarly trivial for in-client support.

markdroth: It's not quite as trivial for gRPC, because we would want to not actually create the child policies for any targets that we're not actually going to talk to, so there'd be a few different code-paths where we'd need conditionals for this.

This is certainly not insurmountable, but I'd prefer to avoid the extra complexity if we can avoid it.

grpc/lookup/v1/rls.proto
```diff
@@ -38,7 +38,14 @@ message RouteLookupRequest {
 message RouteLookupResponse {
   // Actual addressable entity to use for routing decision, using syntax
   // requested by the request target_type.
+  // This field is deprecated in favor of the new "targets" field, below.
   string target = 1;
```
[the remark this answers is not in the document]
cdmetcalf: Done.

grpc/lookup/v1/rls_config.proto
cdmetcalf: Thinking about it some more, I do like the idea of using x-google-rls-data to send dark launch information to the backend service.  This also lets them handle questions of different requests having different weights (CPU or disk intensive etc).  I've reverted the original commit.

PR #78, opened 2020-03-30
load_balancer: Remove never-used load_balancer_delegate

ejona86:
It was part of the initial design but then scrapped early in grpclb's
life. No client has supported it. Remove it to prevent confusion.

@vishalpowar, would you confirm my understanding is correct, that
nothing uses this? I didn't see any users internally.

CC @wicharypawel


ejona86:
@vishalpowar, ping

PR #80, opened 2020-04-27
Add examples/helloworld.proto

chalin:
As discussed with @ejona86.

The version submitted here was copied from the `grpc-go` repo: [grpc-go/examples/helloworld/helloworld/helloworld.proto](https://github.com/grpc/grpc-go/blob/master/examples/helloworld/helloworld/helloworld.proto). No changes have been made to this copy.

Eric, we had discussed possible changes such as: adding a canonical link as a comment, and possibly additional language options. Would you like those added to this PR or a followup PR?

cc @dfawley @menghanl

### DIFF
--- a/grpc/examples/helloworld.proto
+++ b/grpc/examples/helloworld.proto
@@ -1,0 +1,37 @@
+// Copyright 2015 gRPC authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


chalin:
@ejona86:
- Should we consider adding the "streaming" version of hello world, given here: [examples/src/main/proto/hello_streaming.proto](https://github.com/grpc/grpc-java/blame/master/examples/src/main/proto/hello_streaming.proto)? 
- The gRPC quick starts walk the reader through an update to the `helloworld.proto` file -- the addition of the `Greeter.SayHelloAgain()` method. Should we consider adding that modified version of the `.proto` file here too?

ejona86:
@chalin, let's not worry about hello_streaming.proto . IIRC, that is java-specific.

I don't think we need to deal with `Greeter.SayHelloAgain()` here. There's not much need for compatibility there, and the user would be doing the modification themselves anyway.

PR #82, opened 2020-05-01
README.md: Expand the expected usage description

ejona86:
This also removes the expectation of the sanity tests, mainly because
they never were created and were canned before being created because
they were originally invisioned to be CI tests, which would fail PRs the
moment a comment was made to grpc-proto.

We are instead using scripts and other solutions to keep the chances of
manual changes low, while only using a single copy of the protos
in Google's internal build. Since repositories sync to Google-internal
frequently, manual changes that "get through the cracks" and break
functionality should still be discovered promptly.

Not all of that is in the README, though, because it proved hard to
integrate such that the prose was still short and to the point, while
also not being confusing to non-Google-devs. And I don't actually expect
many Google-devs would read it anyway...

CC @chalin

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ peripheral services around gRPC such as health checking, load balancing etc..
 
 ## Usage
 
-For Bazel users, it can included directly as an http_repository.
+For Bazel users, it can be included directly as an http_repository.
 
 For others it is expected the proto files will be copied to other gRPC
 repositories as needed (e.g., `grpc/grpc`, `grpc/grpc-go`, etc.). However, those
--- a/README.md
+++ b/README.md
@@ -1,13 +1,21 @@
 # grpc-proto
-This repository contains common protocol definitions for peripheral services
-around gRPC such as health checking, load balancing etc..
+This repository contains the canonical common protocol definitions for
+peripheral services around gRPC such as health checking, load balancing etc..
 
-## Planned Usage
+## Usage
 
-For ease of development, proto files will still be copied to the other gRPC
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,13 +1,21 @@
...
+peripheral services around gRPC such as health checking, load balancing etc..
 
-## Planned Usage
+## Usage
 
-For ease of development, proto files will still be copied to the other gRPC
-repositories (e.g., `grpc/grpc`, `grpc/grpc-go`, etc.). Sanity tests will be
-added to verify that common proto files match the "ground truth" files contained
-here.
+For Bazel users, it can included directly as an http_repository.
```
sanjaypujare, line 7: s/can included/can be included/

README.md
```diff
@@ -1,13 +1,21 @@
...
-## Planned Usage
+## Usage
 
-For ease of development, proto files will still be copied to the other gRPC
-repositories (e.g., `grpc/grpc`, `grpc/grpc-go`, etc.). Sanity tests will be
-added to verify that common proto files match the "ground truth" files contained
-here.
+For Bazel users, it can be included directly as an http_repository.
+
+For others it is expected the proto files will be copied to other gRPC
```
chalin, line 9: How about:
> Other gRPC repos, and non-Bazel users are expected to copy the  proto files from this repo as needed. However, copied files should not be modified. Changes should be made to proto files in this repo before being recopied. This prevents ...

ejona86: I didn't take this as-is, but integrated a lot of what it suggested.

PR #83, opened 2020-05-15
Clean up unused fields in grpc/lookup/v1 protos

cdmetcalf:
The RouteLookupResponse.target field was deprecated and is no longer
in use by any server or client, so remove entirely.

The RouteLookupConfig.request_processing_strategy was likewise
deprecated when ASYNC_LOOKUP_DEFAULT_TARGET_ON_MISS stopped being
supported and it became equivalent to just a presence check on
default_target, so remove it now.

### DIFF
--- a/grpc/lookup/v1/rls.proto
+++ b/grpc/lookup/v1/rls.proto
@@ -36,20 +36,17 @@ message RouteLookupRequest {
 }
 
 message RouteLookupResponse {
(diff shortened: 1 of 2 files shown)


ejona86:
We shouldn't merge this until the import would be okay.

creamsoup:
> LGTM
> FYI java RLS just received a change to not use those fields. import may needs to wait few days.

the change is merged.

PR #91, opened 2021-02-02
Configure Github Actions and delete Travis

dfawley:


.github/workflows/testing.yml
```diff
@@ -0,0 +1,25 @@
+# This is a basic workflow to help you get started with Actions
+
+name: Testing
+
+# Controls when the action will run. 
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: 0 0 * * * # daily at 00:00
```
ejona86, line 10: Do we need this aggressive of a schedule? I'd be happy without a schedule, because I don't think anyone will look at the results.

ejona86 (approved):

PR #94, opened 2021-06-18
Revert "Add SDK Authorization Policy proto."

ashithasantosh:
Reverts grpc/grpc-proto#88

gRPC Authorization API will be consuming proto in JSON format. Also, the proto is not used internally for implementation.


ejona86 (approved):

ejona86:
Appears all usages of this have already been cleaned up. Internally this was already removed, and I don't see any usages in grpc, grpc-java, or grpc-go repos.

PR #101, opened 2022-01-11
more LB policy config updates

markdroth:
- Add `XdsServer` message and use it in new fields in `xds_cluster_resolver_experimental` and `xds_cluster_impl_experimental` configs, as per the latest draft of grpc/proposal#268
- Deprecate `eds_experimental` policy config, as per [gRFC A37](https://github.com/grpc/proposal/blob/master/A37-xds-aggregate-and-logical-dns-clusters.md)
- Add missing config for `xds_cluster_manager_experimental` policy, as per [gRFC A31](https://github.com/grpc/proposal/blob/master/A31-xds-timeout-support-and-config-selector.md)

### DIFF
--- a/grpc/service_config/service_config.proto
+++ b/grpc/service_config/service_config.proto
@@ -236,9 +236,33 @@ message WeightedTargetLoadBalancingPolicyConfig {
   map<string, Target> targets = 1;
 }
 
+// Configuration for xds_cluster_manager_experimental LB policy.
+message XdsClusterManagerLoadBalancingPolicyConfig {
+  message Child {
+    repeated LoadBalancingConfig child_policy = 1;
+  }
+  map<string, Child> children = 1;
+}
+
 // Configuration for the cds LB policy.
 message CdsConfig {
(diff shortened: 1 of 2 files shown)


grpc/service_config/service_config.proto
```diff
@@ -236,11 +236,35 @@ message WeightedTargetLoadBalancingPolicyConfig {
   map<string, Target> targets = 1;
 }
 
+// Configuration for xds_cluster_manager_experimental LB policy.
+message XdsClusterManagerLoadBalancingPolicyConfig {
+  message Child {
+    repeated LoadBalancingConfig child_policy = 1;
```
dfawley, line 242: Should this `json_name` to `child_policy`?

dfawley (approved):

grpc/service_config/service_config.proto
markdroth: I don't think so.  We use the normal JSON form (`childPolicy`) for these fields in gRPC.

PR #102, opened 2022-01-20
Add common java options to the testing/*.protos. 

anicr7:
This PR adds common java proto options to the testing files.

Particularly the 3 main options include

* Java multiple files = true
* Java package which is being set to io.grpc.testing
* Java outer classname which is being set to fileNameProto

### DIFF
--- a/grpc/testing/test.proto
+++ b/grpc/testing/test.proto
@@ -23,6 +23,10 @@ import "grpc/testing/messages.proto";
 
 package grpc.testing;
 
+option java_multiple_files = true;
+option java_package = "io.grpc.testing";
+option java_outer_classname = "TestProto";
+
 // A simple service to test the various types of RPCs and experiment with
 // performance with various types of payload.
 service TestService {
--- a/grpc/testing/test.proto
+++ b/grpc/testing/test.proto
@@ -23,9 +23,7 @@ import "grpc/testing/messages.proto";
 
 package grpc.testing;
 
-option java_multiple_files = true;
-option java_package = "io.grpc.testing";
-option java_outer_classname = "TestProto";
+option java_package = "io.grpc.testing.integration";
 
 // A simple service to test the various types of RPCs and experiment with
 // performance with various types of payload.
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -28,10 +28,22 @@ proto_library(
     srcs = ["grpc/testing/messages.proto"],
 )
 
+java_proto_library(
+    name = "messages_java_proto",
+    visibility = ["//visibility:public"],
+    deps = [":messages_proto"],
+)
+
 proto_library(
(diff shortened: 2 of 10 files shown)


grpc/testing/test.proto
```diff
@@ -23,6 +23,10 @@ import "grpc/testing/messages.proto";
 
 package grpc.testing;
 
+option java_multiple_files = true;
+option java_package = "io.grpc.testing";
+option java_outer_classname = "TestProto";
```
dapengzhang0, line 28: The java options for `test.proto`, `messages.proto`, and `empty.proto` should be the same as https://github.com/grpc/grpc-java/tree/v1.43.2/interop-testing/src/main/proto/grpc/testing that java library is currently using, because these have been "public" APIs. The `java_package` option in other files should follow suite to be `io.grpc.testing.integration`.

anicr7: Penn, 

The proto across java and this package seem to be very slightly different. For example messages.proto here is


// The type of payload that should be returned.
enum PayloadType {
  // Compressable text format.
  COMPRESSABLE = 0;
}
// A block of data, to simply increase gRPC message size.
message Payload {
  // The type of data in body.
  PayloadType type = 1;
  // Primary contents of payload.
  bytes body = 2;
}

where as in the java grpc/grpc-java@v1.43.2/interop-testing/src/main/proto/grpc/testing has

// A block of data, to simply increase gRPC message size.
message Payload {
  reserved 1;

  // Primary contents of payload.
  bytes body = 2;
}


Is it ok if they have the same package name and Class name if they are slightly different?

dapengzhang0: hmmm , seems only Java repo made the change years ago, anyone in C and go knows about the change?
@yashkt @apolcyn @menghanl

menghanl: We don't keep copies of these `testing/*.proto` files in go. So whatever is in the grpc-proto repo is what we use.

yashykt: I wasn't part of the discussion to deprecate it, but I do some tests using it.

anicr7: @dapengzhang0,

I am trying to think of the best way we can converge here. As these fields are still used in Go/C++, I don't think we can remove it.

Can we submit this into a different java package and then slowly move the Java tests to use this proto instead?

Let me know if you have a better idea in mind.

dapengzhang0: I don't care about the difference in the deprecated field, we can leave it as-is for now. But the proto options for java should be the same as what has already been used by grpc-java now.

anicr7: Changed the proto options to match the options in grpc/grpc-java@v1.43.2/interop-testing/src/main/proto/grpc/testing .

dapengzhang0 (approved):

PR #104, opened 2022-02-14
Add outlier_detection LB policy config

murgatroid99:
These changes are described in grpc/proposal#281.

### DIFF
--- a/grpc/service_config/service_config.proto
+++ b/grpc/service_config/service_config.proto
@@ -611,7 +611,7 @@ message LoadBalancingConfig {
     XdsConfig xds_experimental = 5 [json_name = "xds_experimental",
                                     deprecated = true];
 
-    // Next available ID: 14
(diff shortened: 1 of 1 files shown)


grpc/service_config/service_config.proto
```diff
@@ -484,6 +576,8 @@ message LoadBalancingConfig {
 
     RoundRobinConfig round_robin = 1 [json_name = "round_robin"];
 
+    OutlierDetectionLoadBalancingConfig outlier_detection = 14 [json_name = "outlier_detection"]
```
markdroth, line 579: Please update the "next available ID" comment on line 614 below.

markdroth (approved):
LGTM, modulo one small nit.

grpc/service_config/service_config.proto
murgatroid99: Done.

PR #109, opened 2022-06-17
add orca_oob_report to message.proto stremaingoutputcallrequest

YifeiZhuang:
discussed in https://github.com/grpc/grpc-java/pull/9260


ejona86:
We plan to delete `SimpleRequest.orca_oob_report`, but we need the new field to convert Java over first. Once Java is fixed, we'll delete the old field.

ejona86 (approved):

PR #110, opened 2022-06-23
Update control.proto to be in sync with the one in grpc/grpc.

wanlin31:
The control.proto in grpc/grpc was updated by 
https://github.com/grpc/grpc/pull/29207. This pr updates
the control.proto file in this repo to be in sync with
control.proto in grpc/grpc.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -53,7 +53,10 @@ proto_library(
         "grpc/testing/payloads.proto",
         "grpc/testing/stats.proto",
     ],
-    deps = [":stats_proto"],
+    deps = [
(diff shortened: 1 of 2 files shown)


apolcyn (approved):

ejona86:
Depending on well-known protos may cause problems. We had to avoid Empty to avoid issues in the distant past. Be aware this may need to be rolled back, and thus rolled back in grpc/grpc. The protos really need to change here first.

This change does not currently compile.

ejona86 (approved):
Import this into google3 after it is merged.

PR #115, opened 2022-11-09
add RLS LB policy config

markdroth:
Also a couple of other cleanups:
- Add the missing `_experimental` suffix to the outlier detection config.
- Reorganize the fields a bit to be under the right comment headings.
- Add comments on the ring_hash min/max ring size fields matching the behavior described in https://github.com/grpc/proposal/pull/338.

### DIFF
--- a/grpc/service_config/service_config.proto
+++ b/grpc/service_config/service_config.proto
@@ -34,6 +34,7 @@ import "google/protobuf/duration.proto";
 import "google/protobuf/struct.proto";
 import "google/protobuf/wrappers.proto";
 import "google/rpc/code.proto";
+import "grpc/lookup/v1/rls_config.proto";
 
 option java_package = "io.grpc.serviceconfig";
 option java_multiple_files = true;
@@ -328,6 +329,18 @@ message WeightedTargetLoadBalancingPolicyConfig {
   map<string, Target> targets = 1;
 }
 
+// Config for RLS LB policy.
+message RlsLoadBalancingPolicyConfig {
+  // Route lookup config.
+  grpc.lookup.v1.RouteLookupConfig route_lookup_config = 1;
+  // Service config to use for the RLS channel.
+  ServiceConfig route_lookup_channel_service_config = 2;
+  // Child policy config.
+  repeated LoadBalancingConfig child_policy = 3;
+  // Field name to add to child policy config to contain the target name.
+  string child_policy_config_target_field_name = 4;
+}
+
 // Configuration for xds_cluster_manager_experimental LB policy.
 message XdsClusterManagerLoadBalancingPolicyConfig {
   message Child {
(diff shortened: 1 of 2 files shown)


grpc/service_config/service_config.proto
```diff
@@ -328,6 +329,18 @@ message WeightedTargetLoadBalancingPolicyConfig {
   map<string, Target> targets = 1;
 }
 
+// Config for RLS LB policy.
+message RlsLoadBalancingPolicyConfig {
+  // Route lookup config.
```
dfawley, line 334: IMO it's better to not have a comment then to have a comment that doesn't provide any additional information.

grpc/service_config/service_config.proto
```diff
@@ -328,6 +329,18 @@ message WeightedTargetLoadBalancingPolicyConfig {
...
+  grpc.lookup.v1.RouteLookupConfig route_lookup_config = 1;
+  // Service config to use for the RLS channel.
+  ServiceConfig route_lookup_channel_service_config = 2;
+  // Child policy config.
```
dfawley, line 338: Similar.

grpc/service_config/service_config.proto
```diff
@@ -607,8 +624,12 @@ message LoadBalancingConfig {
         [json_name = "xds_cluster_resolver_experimental"];
     XdsClusterImplLoadBalancingPolicyConfig xds_cluster_impl_experimental = 12
         [json_name = "xds_cluster_impl_experimental"];
+    XdsWrrLocalityLoadBalancingPolicyConfig xds_wrr_locality_experimental = 16
+        [json_name = "xds_wrr_locality_experimental"];
```
dfawley, line 628: I don't really mind, but why intentionally order this way vs. by tag #?

grpc/service_config/service_config.proto
markdroth: Done.

grpc/service_config/service_config.proto
markdroth: Done.

grpc/service_config/service_config.proto
markdroth: It was incorrect to put the `xds_wrr_locality` and `least_request` policies under the "Deprecated xDS-related policies" comment on line 634.  These are not deprecated.

It also makes sense to put the `outlier_detection` under the "REMAINING POLICIES ARE EXPERIMENTAL" comment on line 608.

dfawley: Oh I mean _within this section_, why not order by tag #?  This one entry (16) was added between 12 and 13, and not next to 17, all in this section.

markdroth: Ah, okay -- sorry, I misunderstood your question.

I'm ordering all of the xDS policies here by their position in the LB policy tree, from top to bottom.

dfawley: Seems reasonable to me.

dfawley (approved):

PR #116, opened 2022-12-14
service config: add WRR LB policy config

markdroth:


grpc/service_config/service_config.proto
```diff
@@ -196,6 +196,21 @@ message PickFirstConfig {}
 // Configuration for round_robin LB policy.
 message RoundRobinConfig {}
 
+// Configuration for weighted_round_robin LB policy.
```
yousukseung, line 199: For the same reason we prefixed it with "client_side" in Envoy, can we mention it's 'client-side' in the comment here?

markdroth: Not necessary in this case, because this is the only WRR policy that gRPC plans to support.

yousukseung (approved):

grpc/service_config/service_config.proto
```diff
@@ -202,13 +202,18 @@ message WeightedRoundRobinLbConfig {
   // the endpoints. By default, per-request utilization reporting is used.
   google.protobuf.BoolValue enable_oob_load_report = 1;
 
+  // Load reporting interval to request from the server.  Note that the
+  // server may not provide reports as frequently as the client requests.
+  // Used only when enable_oob_load_report is true.
```
yousukseung: "Defaults to 10 seconds."?

markdroth: Yup, already did that -- you must not have been looking at the latest commit.

PR #118, opened 2023-02-13
Add a stale_if_error field to the RouteLookupConfig proto

imyifanz:
Add a stale_if_error field to the RouteLookupConfig proto to allow stale entries when RLS service has an error


imyifanz:
Closing this request since the code change is no longer needed.

PR #123, opened 2023-04-04
.github/workflows: Bump Bazel for testing to 5.0.0

ejona86:
Now you can run locally using the version of your choice.


ejona86:
@dfawley, friendly ping.

dfawley (approved):
Surely there's a newer version by now?

ejona86:
Yeah, but for compat reasons I was still planning to upgrade to this older version first (even though I forgot about this PR and was creating a new PR, and then noticed I already had it).

ejona86:
(And although there are newer versions, we tend to support the two most recent versions and Bazel 7 dropped just today.)

PR #132, opened 2023-08-17
Improvements for metadata matching

eugeneo:
1. Add mention of the "*" matching.
2. Return entry type (Initial/Trailing)
3. Reshuffled some comments.

### DIFF
--- a/grpc/testing/messages.proto
+++ b/grpc/testing/messages.proto
@@ -1,4 +1,3 @@
-
 // Copyright 2015-2016 gRPC authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


grpc/testing/messages.proto
```diff
@@ -207,20 +206,29 @@ message LoadBalancerStatsRequest {
...
     string key = 1;
+    // Value, exactly as received from the server.
     string value = 2;
+    // Metadata type is either Initial or Trailing
+    MetadataType type = 3;
   }
   message RpcMetadata {
     // metadata values for each rpc for the keys specified in
     // LoadBalancerStatsRequest.metadata_keys.
-    // metadata keys and values are returned exactly as was received
```
sanjaypujare, line 222: Why is this comment removed? Why is it no longer true (that "metadata keys and values are returned exactly as was received")?

grpc/testing/messages.proto
```diff
@@ -207,20 +206,29 @@ message LoadBalancerStatsRequest {
...
-  // response header+trailer we want the values of
+  // Response header + trailer metadata entries we want the values of.
+  // Matching of the keys is case-insensitive as per rfc7540#section-8.1.2
+  // * (asterisk) is a special value that will return all metadata entries
   repeated string metadata_keys = 3;
 }
 
 message LoadBalancerStatsResponse {
+  enum MetadataType {
+    Initial = 0;
```
sanjaypujare, line 217: This is the default so if the type is not present it is assumed to be "initial" right?

sanjaypujare (approved):
Looks good but I have a couple of clarifying questions. 

Also wouldn't it be better to make the change first in grpc-proto copy?

grpc/testing/messages.proto
eugeneo: I moved the comments above, to metadata entry. I also elaborated there that the key has same case as server supplied, and not the value from the original request. I decided not to duplicate the comment so there's no unsyncing in the future.

grpc/testing/messages.proto
eugeneo: Yes.

PR #133, opened 2023-08-23
Add application-specific optional `extensions` fields to rls.proto

efimki:
Add  application-specific optional `extensions` fields to `RouteLookupRequest` and `RouteLookupResponse` proto messages.

### DIFF
--- a/grpc/lookup/v1/rls.proto
+++ b/grpc/lookup/v1/rls.proto
@@ -16,6 +16,8 @@ syntax = "proto3";
 
 package grpc.lookup.v1;
 
+import "google/protobuf/any.proto";
+
 option go_package = "google.golang.org/grpc/lookup/grpc_lookup_v1";
 option java_multiple_files = true;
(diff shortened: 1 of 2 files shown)


grpc/lookup/v1/rls.proto
```diff
@@ -37,6 +37,8 @@ message RouteLookupRequest {
   string stale_header_data = 6;
   // Map of key values extracted via key builders for the gRPC or HTTP request.
   map<string, string> key_map = 4;
+  // Application-specific optional extensions.
+  repeated Any extensions = 5;
```
markdroth, line 41: These fields need to be of type "google.protobuf.Any" instead of just "Any".

Also, you'll need to add the following around line 19 above:

```
import "google/protobuf/any.proto";
```

And you'll need to add the appropriate dependency in the BUILD.bazel file.

efimki: Thanks for the suggestions! Done.

markdroth (approved):

PR #134, opened 2023-08-28
Address protobuf linter warnings.

eugeneo:
1. Default value should not be a value that is in use.
2. Make enum nested so enum values are not global.


ejona86 (approved):

sanjaypujare (approved):
LGTM

Issue #127, opened 2023-06-13
Add idempotency_level to health check schema


akshayjshah:
What do you think of adding [`idempotency_level`](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/descriptor.proto#L815) options to the [health check schema](https://github.com/grpc/grpc-proto/blob/master/grpc/health/v1/health.proto)? The most appropriate level seems to be `NO_SIDE_EFFECTS`.

Even though gRPC always uses HTTP POST, these annotations would be useful:

* They document the expected behavior of servers.
* They allow client-side interceptors to adopt more aggressive retry policies.
* They allow gRPC-aware middleboxes to cache responses (though setting a TTL may be tricky).
* They allow other RPC protocols to serve the `Check` RPC using HTTP `GET`s.

Happy to open a PR if there's any interest!

akshayjshah:
@markdroth Any chance you've got a moment to consider this? (Tagging you because you're the last person to have touched `health.proto`.)

markdroth:
I don't think we want this, because we don't want health statuses being cached -- the whole point is to verify that the server is healthy, and if it goes unhealthy, we don't want the client still getting a cached signal that it was healthy.

ejona86:
We don't respond with any caching headers, so they wouldn't be cached unless someone went out of their way to force caching in a proxy or using an interceptor to add the appropriate response headers.

markdroth:
True, but I'm a little leery of making a change that might encourage doing something like that.  And it's really not clear to me that there's any other benefit here anyway.  I don't think we want retries of these RPCs -- if a server is too busy to serve a health check, it should be considered unhealthy.  And this is a gRPC-specific health check mechanism; I don't see a need to serve such requests via any other protocol.

ejona86:
I'd expect there's _someone_ that wants this for GET to point an HTTP health checker at. But I would agree insofar someone should tell us their specific (legitimate) usecase and then we could add it.

akshayjshah:
> True, but I'm a little leery of making a change that might encourage doing something like that.

I can understand some apprehension about seeming to encourage caching or retries. But zooming out a bit: why is gRPC more susceptible to these problems than other HTTP APIs? Nearly all HTTP health checks are `GET`s, and other systems aren't _unexpectedly_ beset by stale check results or fruitless retries. `GET`s give engineers more options when unusual situations come up. Marking `Check` side effect free offers those same options to engineers using gRPC.

[...]

PR #135, opened 2023-09-08
Improve documentation comments for health service

akshayjshah:
This commit somewhat improves the documentation of `grpc.health.v1.Health`:

- It directs readers to [the specification](https://github.com/grpc/grpc/blob/master/doc/health-checking.md), which feels a bit too long to inline into comments.
- It explains the semantics of the `Check` RPC a bit more thoroughly, clarifying that implementations should be idempotent and side effect free.

Hopefully, this PR faithfully follows the discussion with @ejona86 in #127.

### DIFF
--- a/grpc/health/v1/health.proto
+++ b/grpc/health/v1/health.proto
@@ -39,9 +39,19 @@ message HealthCheckResponse {
   ServingStatus status = 1;
 }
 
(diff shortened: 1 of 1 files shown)


ejona86 (approved):

markdroth (approved):

akshayjshah:
Thanks for the quick review!

PR #136, opened 2023-09-19
[xds test] Add ReturnAll method to a HookServer

eugeneo:
I propose to add a ReturnAll method to a Hook server. This replaces the HookRequest with the Return method. Ultimately, this would allow us to completely remove the SendHookRequest method from the XdsUpdateHealthService when the tests no longer rely on it.

### DIFF
--- a/grpc/testing/messages.proto
+++ b/grpc/testing/messages.proto
@@ -319,6 +319,14 @@ message TestOrcaReport {
   map<string, double> utilization = 4;
 }
 
+// Request to a ReturnAll method of the HookService. Specifies the status
+// that will be returned to all pending hook requests as well as hook requests
+// that will arrive after this call.
+message ReturnAllRequest {
+  int32 grpc_code_to_return = 1;
+  string grpc_status_description = 2;
+}
+
 message HookRequest {
   enum HookRequestCommand {
     // Default value
--- a/grpc/testing/messages.proto
+++ b/grpc/testing/messages.proto
@@ -319,10 +319,8 @@ message TestOrcaReport {
   map<string, double> utilization = 4;
 }
 
-// Request to a ReturnAll method of the HookService. Specifies the status
-// that will be returned to all pending hook requests as well as hook requests
-// that will arrive after this call.
-message ReturnAllRequest {
+// Status that will be return to callers of the Hook method
+message SetReturnStatusRequest {
   int32 grpc_code_to_return = 1;
   string grpc_status_description = 2;
 }
--- a/grpc/testing/test.proto
+++ b/grpc/testing/test.proto
@@ -91,8 +91,7 @@ service LoadBalancerStatsService {
       returns (LoadBalancerAccumulatedStatsResponse) {}
 }
 
-// Hook service. Used to keep a connection alive so Kubernates does not shut
-// down the pod.
+// Hook service. Used to keep Kubernetes from shutting the pod down.
 service HookService {
   // Sends a request that will "hang" until the return status is set by a call
   // to a SetReturnStatus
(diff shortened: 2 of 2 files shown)


gnossen (approved):
Thanks for the change Eugene! Please get Sanjay's approval as well.

grpc/testing/test.proto
```diff
@@ -95,6 +95,7 @@ service LoadBalancerStatsService {
 // with HookRequestCommand::START
 service HookService {
   rpc Hook(grpc.testing.Empty) returns (grpc.testing.Empty);
+  rpc ReturnAll(ReturnAllRequest) returns (grpc.testing.Empty);
```
sanjaypujare, line 98: Why is this needed? This is different from what was in the design.

grpc/testing/messages.proto
```diff
@@ -319,6 +319,14 @@ message TestOrcaReport {
   map<string, double> utilization = 4;
 }
 
+// Request to a ReturnAll method of the HookService. Specifies the status
+// that will be returned to all pending hook requests as well as hook requests
+// that will arrive after this call.
+message ReturnAllRequest {
```
sanjaypujare, line 325: This message does not seem to serve any purpose

grpc/testing/test.proto
eugeneo: I feel like it is more logical. Current plan is to move the hook service to the maintenance port and have it run all the time. This makes START and STOP hook requests obsolete, leaving only the Return one. I really don't see a need to couple XdsHealthService to the HookService just to forward the Return command.

sanjaypujare: Okay, discussed offline. The Hook command function and the actual Hook function are combined into a single service. Sounds good.

grpc/testing/messages.proto
eugeneo: It is the status that will be returned to hook requests, same as https://github.com/grpc/grpc-proto/blob/master/grpc/testing/messages.proto#L334

grpc/testing/test.proto
sanjaypujare: But what about the `reset` functionality described in the doc? Since `ReturnAll` also applies to future `Hook` calls if there is a need to reset this state we should have it, no?

grpc/testing/test.proto
```diff
@@ -95,6 +95,7 @@ service LoadBalancerStatsService {
 // with HookRequestCommand::START
```
sanjaypujare, line 95: Since the HookService serves both the `Hook` calls and the command call (like `ReturnAll`) there is no connection with the HealthService any more. You should update the comment on line 94

sanjaypujare (changes requested):
Pls add the `reset` call too

grpc/testing/test.proto
eugeneo: Done

grpc/testing/test.proto
eugeneo: Added ClearReturnStatus

eugeneo (commented):
@sanjaypujare I added ClearReturnStatus, renamed setter and added some comments. PTAL.

grpc/testing/test.proto
```diff
@@ -91,10 +91,16 @@ service LoadBalancerStatsService {
       returns (LoadBalancerAccumulatedStatsResponse) {}
 }
 
-// Hook service that may be started on request by calling XdsUpdateHealthService
-// with HookRequestCommand::START
+// Hook service. Used to keep a connection alive so Kubernates does not shut
```
sanjaypujare, line 94: nit: "kubernetes"

Also nit: there is no connection to keep alive. In the current case it is used to keep the pod alive for currently assigned sessions to the pod.

sanjaypujare (approved):
2 nits found otherwise looks good

grpc/testing/test.proto
eugeneo: Addressed. Thank you for the review!

PR #138, opened 2023-12-11
Add network_latency_ms field to ALTS's NextHandshakeMessageReq.

matthewstevenson88:
Already internally-reviewed (see cl/589894002 and cl/589974172).

### DIFF
--- a/grpc/gcp/handshaker.proto
+++ b/grpc/gcp/handshaker.proto
@@ -156,6 +156,11 @@ message NextHandshakeMessageReq {
   // that the peer's out_frames are split into multiple NextHandshakerMessageReq
   // messages.
   bytes in_bytes = 1;
+
+  // Number of milliseconds between when the application send the last handshake
+  // message to the peer and when the application received the current handshake
+  // message (in the in_bytes field) from the peer.
+  optional uint32 network_latency_ms = 2;
 }
 
 message HandshakerReq {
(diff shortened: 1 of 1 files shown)


ejona86:
(FWIW, it tends to be useful to copy the CL number here so we can follow things a bit easier. Especially further in the future and trying to figure out what we did.)

ejona86:
`optional` is not allowed? We should able able to do that, although we may find some places that need upgrades. It looks like the CI here is just using too old a protobuf.

`optional` in proto3 provides presence information, and it looks like you would want that. How much do you care?

matthewstevenson88:
> `optional` is not allowed? We should able able to do that, although we may find some places that need upgrades. It looks like the CI here is just using too old a protobuf.
> 
> `optional` in proto3 provides presence information, and it looks like you would want that. How much do you care?

It's not a big deal in this case, because I only care about the positive values anyway, so "zero" and "not set" are equivalent.

If there's an easy fix to allow `optional` so that the next person doesn't run into this issue (e.g. likely just update the @com_google_protobuf dep to the newest version?), I'm happy to do it in this PR or in a follow-up, whatever you prefer.

ejona86:
> I'm happy to do it in this PR or in a follow-up, whatever you prefer.

Thanks for the offer, but I think I have it. We're going to need to merge https://github.com/grpc/grpc-proto/pull/123 and then bumping past 21.x required a non-obvious line and bumping to 25 required further changes to bazel. I've fought this before in other circumstances.

ejona86 (approved):

matthewstevenson88:
Thanks Eric!

PR #140, opened 2023-12-19
service config: update xDS LB policy configs for gRFC A74

markdroth:
See grpc/proposal#404.

### DIFF
--- a/grpc/service_config/service_config.proto
+++ b/grpc/service_config/service_config.proto
@@ -317,7 +317,7 @@ message OutlierDetectionLoadBalancingConfig {
 
   // If set, success rate ejections will be performed
   SuccessRateEjection success_rate_ejection = 5;
-  
+
   // If set, failure rate ejections will be performed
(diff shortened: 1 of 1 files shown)


grpc/service_config/service_config.proto
```diff
@@ -412,6 +415,7 @@ message XdsServer {
 }
 
 // Configuration for xds_cluster_resolver LB policy.
+// Deprecated.
 message XdsClusterResolverLoadBalancingPolicyConfig {
```
dfawley, line 419: Optional: It might be nice to move these deprecated proto messages to another file so they don't clutter up the non-deprecated ones.  E.g. `deprecated.proto`?

dfawley (approved):
LGTM; please see optional comment inline.

grpc/service_config/service_config.proto
markdroth: I've moved the deprecated messages to a separate section at the end of the file.

PR #142, opened 2024-02-07
add channel opts, rpc error, additional state to channelz

yousukseung:


yousukseung:
@markdroth kindly ping.

PR #144, opened 2024-02-23
chore(bazel): add MODULE.bazel files for bzlmod

mmorel-35:
Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,0 +1,15 @@
+module(
+    name = "grpc_proto",
+    version = "0.0.0",
+    repo_name = "io_grpc_grpc_proto",
+)
+
+bazel_dep(name = "rules_proto", version = "4.0.0")
+bazel_dep(name = "protobuf", version = "23.1", repo_name = "com_google_protobuf")
+# -- bazel_dep definitions -- #
+
+http_archives_extension = use_extension("//bazel:http_archives_extension.bzl", "http_archives_extension")
+use_repo(http_archives_extension, "com_google_googleapis")
+
+switched_rules_extension = use_extension("//bazel:switched_rules_extension.bzl", "switched_rules_extension")
+use_repo(switched_rules_extension, "com_google_googleapis_imports")
--- a/bazel/switched_rules_extension.bzl
+++ b/bazel/switched_rules_extension.bzl
@@ -1,0 +1,8 @@
+load("@com_google_googleapis//:repository_rules.bzl", "switched_rules_by_language")
+
+def _switched_rules_extension_impl(ctx):
+    switched_rules_by_language(
+        name = "com_google_googleapis_imports",
+    )
+
+switched_rules_extension = module_extension(implementation = _switched_rules_extension_impl)
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -7,16 +7,18 @@ on:
 permissions:
   contents: read
 
-env:
-  USE_BAZEL_VERSION: 5.0.0
-
 jobs:
   test:
     runs-on: ubuntu-latest
-
+    env:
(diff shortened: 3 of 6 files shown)


.bazeliskrc
```diff
@@ -0,0 +1 @@
+USE_BAZEL_VERSION=7.0.0
```
ejona86, line 1: We don't want to force the version. Generally locally we use the latest Bazel, but then the CI uses the oldest bazel version we support. Let's remove this and keep the github action specifying the version to use.

We also still support Bazel 6, so we want the CI to use Bazel 6.

.bazelrc
```diff
@@ -1 +1,4 @@
+# Enable Bzlmod for every Bazel command
+common --enable_bzlmod
```
ejona86, line 2: You didn't remove `WORKSPACE`, so if I understand correctly bzlmod-enabled builds will use these new files and non-bzlmod builds will use the old files. Seems we need to add another build to the CI that disables bzlmod.

MODULE.bazel
```diff
@@ -0,0 +1,17 @@
+module(
+    name = "grpc_proto",
+    version = "0.0.0",
+    repo_name = "io_grpc_grpc_proto",
+)
+
+bazel_dep(name = "rules_proto", version = "6.0.0-rc2")
+bazel_dep(name = "protobuf", version = "23.1", repo_name = "com_google_protobuf")
```
ejona86, line 8: Can we not use 24.4 as used in WORKSPACE today?

bazel/switched_rules_extension.bzl
```diff
@@ -0,0 +1,149 @@
+load("@com_google_googleapis//:repository_rules.bzl", "switched_rules")
+# -- load statements -- #
+
+def _switched_rules_extension_impl(ctx):
+    switched_rules(
```
ejona86, line 5: Why not use `switched_rules_by_language` like we used to do? This seems brittle.

.bazeliskrc
mmorel-35: I have adapted the workflow to test v6 and v7

bazel/switched_rules_extension.bzl
mmorel-35: It's done now

MODULE.bazel
mmorel-35: I'm afraid, it's not available on BCR

ejona86: 23.1 was released May 2023. 23.2 was released the same month. 24.0 was released Aug 2023. And 25.0 was released Nov 2023. Looks like 23.1 was added to BCR at the very, very end of 2023.

Either BCR needs updating or we define the module locally (however that should be done). I can accept the skew, but we need to be _able to_ choose something recent, so having this problem out-of-the-gate makes me think we need to go through that process now so we know how it is done.

mmorel-35: I'v tried importing it  but then it requires a lot more bazel_deps which is quite bothering.
I believe BCR needs to be updated but I don't understand why it is not more for such a crucial dependency.

ejona86: Okay. Well, this PR is still a step in the right direction. We can use the older version for now.

ejona86:
I've remove the branch protection that required the `test` GHA test complete, so that we can make the change here.

ejona86:
No need for both @veblush and @gnossen to review. First one wins.

veblush (approved):

veblush:
Thanks for the PR. Are you going to register this to BCR?

mmorel-35:
> Are you going to register this to BCR?

I don't know how to do this

ejona86:
@mmorel-35, thank you!

PR #146, opened 2024-03-07
Synchronize grpc-proto's copy of handshaker.proto with internal version.

matthewstevenson88:


rockspore (approved):
Thanks, @matthewstevenson88!

rockspore:
@matthewstevenson88 When will this be imported?

PR #147, opened 2024-04-04
sync the handshaker.proto with the internal version

rockspore:


grpc/gcp/handshaker.proto
```diff
@@ -125,11 +125,11 @@ message ServerHandshakeParameters {
   // specified. Otherwise, the handshaker chooses a default local identity.
   repeated Identity local_identities = 2;
 
-  // (Optional) An access token created by the caller only intended for use in
-  // ALTS connections. The access token that should be used to authenticate to
-  // the peer. The access token MUST be strongly bound to the ALTS credentials
-  // used to establish the connection that the token is sent over.
-  optional string access_token = 3;
+  // An identity token created by GKE MDS only intended for use in Direct Path
```
matthewstevenson88, line 128: Can we leave the original wording as-is and just omit the word "access"?

matthewstevenson88 (commented):
Thanks for the PR Luwei!

grpc/gcp/handshaker.proto
rockspore: Makes sense. Done.

grpc/gcp/handshaker.proto
```diff
@@ -125,11 +125,11 @@ message ServerHandshakeParameters {
   // specified. Otherwise, the handshaker chooses a default local identity.
   repeated Identity local_identities = 2;
 
-  // (Optional) An access token created by the caller only intended for use in
-  // ALTS connections. The access token that should be used to authenticate to
-  // the peer. The access token MUST be strongly bound to the ALTS credentials
+  // An token created by the caller only intended for use in
```
matthewstevenson88, line 128: nit: s/An/A

matthewstevenson88 (approved):

grpc/gcp/handshaker.proto
rockspore: Oops. Thanks!

PR #148, opened 2024-04-08
fix typo

benjaminp:


ejona86 (approved):
Yeah, we don't want subchannels to be their own father (ancestor). DAG FTW.

PR #153, opened 2024-06-17
use module version of googleapis

mmorel-35:
Notice that the bazel_dep of googleapis is incompatible with bazel 6.


MODULE.bazel
```diff
@@ -4,12 +4,11 @@ module(
...
 # -- bazel_dep definitions -- #
 
-http_archives_extension = use_extension("//bazel:http_archives_extension.bzl", "http_archives_extension")
-use_repo(http_archives_extension, "com_google_googleapis")
-
-switched_rules_extension = use_extension("//bazel:switched_rules_extension.bzl", "switched_rules_extension")
-use_repo(switched_rules_extension, "com_google_googleapis_imports")
+switched_rules = use_extension("@com_google_googleapis//:extensions.bzl", "switched_rules")
+switched_rules.use_languages()
+use_repo(switched_rules, "com_google_googleapis_imports")
```
keith, line 14: looks like we don't even need this? I don't see this referenced anywhere in this repo. does the side effect matter? I don't think it does since the only place com_google_googleapis is used is for a raw proto library, which i think always exists? https://github.com/grpc/grpc-proto/blob/d4056f5f413d38c6e05e1f14f4d208447c0e1673/BUILD.bazel#L244

keith (approved):

MODULE.bazel
ejona86: Yeah, I think you are right. Looking at when it was introduced, it seems it was already not used:
https://github.com/grpc/grpc-proto/commit/e53e1f363c14f30c8ddf42e84ce284da6f6ebff3

I think we'll just use the "if it compiles" approach for confirming this is okay. (The author of that change is around, but it was 4 years ago)

ejona86 (approved):
Given bzlmod users likely want Bazel 7, not supporting Bazel 6 for them seems acceptable. It's not like we've been supporting bzlmod for a long time in this repo. Looks like grpc and grpc-java repos already require Bazel 7 for bzlmod.

MODULE.bazel
```diff
@@ -1,15 +1,10 @@
 module(
     name = "grpc-proto",
     version = "0.0.0",
+    bazel_compatibility = [">=7.0.0"],
```
keith, line 4: unless this is because of this repo, shouldn't this be coming from the transitives? just matters in case the transitives change that requirement

keith (approved):

veblush (approved):

PR #156, opened 2024-07-05
Add bazel targets for s2a proto files

### LINKED ISSUE #155: Unable to compile s2a_context.proto
Similar to #150, there is a syntax error in `grpc/gcp/s2a/s2a_context.proto` (missing semicolon):

```
grpc/gcp/s2a/s2a_context.proto:36:3: Expected ";".
```

pkwarren:
Update the bazel build to add targets for the s2a proto files. Fix syntax error in s2a_context.proto.

Fixes #155.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -247,3 +247,34 @@ proto_library(
         "@com_google_protobuf//:wrappers_proto",
     ],
 )
+
+proto_library(
+    name = "s2a_common_proto",
+    srcs = [
+        "grpc/gcp/s2a/common.proto",
+    ],
+    visibility = ["//visibility:public"],
+)
+
+proto_library(
+    name = "s2a_context_proto",
+    srcs = [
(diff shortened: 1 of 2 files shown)


BUILD.bazel
```diff
@@ -247,3 +247,34 @@ proto_library(
         "@com_google_protobuf//:wrappers_proto",
     ],
 )
+
+proto_library(
```
pkwarren, line 251: Not sure if Java libraries are needed but this at least gets Bazel building the s2a protos.

ejona86: We should be able to remove the java_proto_library from the other ones. No need to add it here.

ejona86:
CC @rmehta19

matthewstevenson88 (approved):
Thanks @ejona86!

ejona86 (approved):

PR #157, opened 2024-07-23
service config: add Random Subsetting LB Policy (gRFC A68)

s-matyukevich:
See https://github.com/grpc/proposal/pull/423 for details.

### DIFF
--- a/grpc/service_config/service_config.proto
+++ b/grpc/service_config/service_config.proto
@@ -476,6 +476,16 @@ message OverrideHostLoadBalancingPolicyConfig {
   repeated HealthStatus override_host_status = 1 [deprecated = true];
 }
 
+// Configuration for the random_subsetting_experimental LB policy.
+message RandomSubsettingLbConfig {
+    // subset_size indicates how many backends every client will be connected to.
+    // The value is required and must be greater than 0.
+    uint32 subset_size = 1;
+    // The config for the child policy.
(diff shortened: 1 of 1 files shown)


grpc/service_config/service_config.proto
```diff
@@ -526,6 +536,8 @@ message LoadBalancingConfig {
     OutlierDetectionLoadBalancingConfig outlier_detection = 15
         [json_name = "outlier_detection_experimental"];
     RlsLoadBalancingPolicyConfig rls = 19 [json_name = "rls_experimental"];
+    RandomSubsettingLbConfig random_subsetting = 21 
+        [json_name = "random_subsetting_experimental"];
```
s-matyukevich, line 540: Should I use `_experimental` suffix here? If yes I may need to update gRFC.

ejona86: Those other experimental policies aren't to be used outside of grpc itself right now. Since yours is for general use, it'd go above the "REMAINING POLICIES ARE EXPERIMENTAL" section and not be experimental. Implementations should keep it _experimental until it is properly tested, but that's separate from this proto.

@dfawley, agreed?

dfawley: Makes sense to me

s-matyukevich: ok, done.

PR #159, opened 2024-08-28
redact token fields in handshaker proto

rockspore:


matthewstevenson88 (approved):

dfawley (approved):

dfawley:
+@ejona86 just in case he actually knows anything at all about these tags, since I don't.

rockspore:
@ejona86 Do you have any comments on this PR? If not, I will merge it. This will have to be manually synced up anyway.

ejona86 (approved):
I found the doc for the option: https://github.com/protocolbuffers/protobuf/blob/1f237eff076d6b658fbefe473ca7542d62bbbdcc/src/google/protobuf/descriptor.proto#L730

PR #164, opened 2024-12-04
s2a: Change S2A proto package name

rmehta19:
After this gets merged, we can modify grpc-java s2a to use the new protos, and then delete the old copy at gcp/s2a/common.proto, gcp/s2a/s2a.proto, gcp/s2a/s2a_context.proto

### DIFF
--- a/grpc/gcp/s2a/proto/v2/common.proto
+++ b/grpc/gcp/s2a/proto/v2/common.proto
@@ -1,0 +1,82 @@
+// Copyright 2024 The gRPC Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// The canonical version of this proto can be found at
+// https://github.com/grpc/grpc-proto/blob/master/grpc/gcp/s2a/common.proto
+
+syntax = "proto3";
+
+package s2a.proto.v2;
+
+option java_multiple_files = true;
+option java_outer_classname = "CommonProto";
+option java_package = "io.grpc.s2a.internal.handshaker";
+
+// The TLS 1.0-1.2 ciphersuites that the application can negotiate when using
+// S2A.
+enum Ciphersuite {
+  CIPHERSUITE_UNSPECIFIED = 0;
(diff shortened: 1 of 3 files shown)


rmehta19:
@matthewstevenson88, @xmenxk

matthewstevenson88 (approved):

grpc/gcp/s2a/proto/v2/common.proto
```diff
@@ -0,0 +1,82 @@
...
+// The canonical version of this proto can be found at
+// https://github.com/grpc/grpc-proto/blob/master/grpc/gcp/s2a/common.proto
+
+syntax = "proto3";
+
+package s2a.proto.v2;
+
+option java_multiple_files = true;
+option java_outer_classname = "CommonProto";
+option java_package = "io.grpc.s2a.internal.handshaker";
```
ejona86, line 24: This conflicts with the internal proto. I think we need to have a better undestanding of the plan.

rmehta19: Thanks for pointing this out Eric. From our email thread about this proto package name change, and what you pointed out about conflicting java package names, my understanding is there are a few steps:
- introduce a new proto here, with the package name `s2a.proto.v2`, with a java package name of `io.grpc.s2a.internal.proto` (or something else).
- once this is merged, delete the [protos](https://github.com/grpc/grpc-java/tree/master/s2a/src/main/proto/grpc/gcp/s2a) in gRPC-Java and replace them with these protos (with a directory structure: s2a/src/main/proto/grpc/gcp/s2a/proto/v2/). This will require importing the new protos from package `io.grpc.s2a.internal.proto` in implementation files (currently the implementation can use protos without importing them because they are in the same java package)
- delete the old protos in this repo
- delete the old protos in gRPC-Java

It would probably be preferable to have the protos continue to live in the same java package as the implementation (@matthewstevenson88, please let me know if it doesn't matter) , in this case, I think we would need to take a different sequence of steps:
- update protos in gRPC-Java to have the package name `s2a.proto.v2`
- send out this PR which creates the new protos with package name `s2a.proto.v2`
- in a followup PR delete the old protos in this package

However I feel like the second process I mentioned might not be feasible, since IIUC, new protos need to be added in this repo before they can be updated in gRPC-Java. 

WDYT? If the first sequence is ok with you and @matthewstevenson88 , we can go ahead and implement it.

grpc/gcp/s2a/proto/v2/common.proto
```diff
@@ -0,0 +1,82 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// The canonical version of this proto can be found at
+// https://github.com/grpc/grpc-proto/blob/master/grpc/gcp/s2a/common.proto
+
+syntax = "proto3";
+
+package s2a.proto.v2;
```
ejona86, line 20: The folder structure needs to match the package name, so this would live in the top-level folder s2a.

PR #168, opened 2025-02-20
s2a: remove s2a protos from grpc-proto

rmehta19:
S2A protos hosted in https://github.com/google/s2a-proto


rmehta19:
@ejona86 , please review, thanks!

ejona86 (approved):

PR #169, opened 2025-04-25
[alts] Update ALTS proto definitions to support ALPN negotiation

anniefrchz:


matthewstevenson88 (approved):

ejona86 (approved):
The formatting for the new messages is surprising, but whatever.

PR #177, opened 2026-06-15
Pin checkout action in testing workflow

Alb3e3:
## Summary

- replace the mutable `actions/checkout@v2` reference in the Testing workflow with the immutable v6.0.3 commit SHA
- keep the existing `contents: read` workflow token restriction unchanged

## Rationale

The testing workflow already runs with a read-only `GITHUB_TOKEN`, but it still loaded `actions/checkout` through a mutable tag. Pinning the action to a full commit SHA reduces supply-chain risk from tag movement or compromised action releases while preserving the existing checkout behavior.

Before the patch, zizmor reported `unpinned-uses` for the checkout step at high confidence. After the patch, the workflow has no medium-or-higher findings under the same scan settings.

## Testing

- `uvx --from zizmor zizmor --min-severity medium --min-confidence medium .github/workflows/testing.yml`
- `git diff --check`

Not run locally: Bazel build, because this container does not have `bazel` or `bazelisk` installed.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -21,7 +21,7 @@ jobs:
             bzlmod: "--enable_bzlmod"
     steps:
       - name: Checkout repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
 
       - name: Run bazel build
         run: bazelisk build ${{ matrix.bzlmod }} //...


ejona86:
Given we have `checkout@2`, we're clearly already having enough trouble upgrading the actions. I don't think we have enough tooling in place to really do this, and the change does make it harder to verify that the SHA is actually what it claims to be. That is, it reduces the risk from ecosystem-scale supply chain attack, but increases the risk from direct supply-chain attack.

Given that the action doesn't have any permissions, the reduced usability doesn't seem to be worth it.

